PR #129, opened 2021-10-05
Base Go CDK Metrics Server Backend

freeqaz:
This commit implements a basic Go Cloud Development Kit backend server
that will eventually become the Metrics Server Backend.

Currently, this service is just a fork of the Guestbook example app.
Fortunately, the example app already has a DB setup along with an HTTP
server to process requests.

I have also added a Readme explaining out metrics server. We should take
that document and make it a part of our Docusaurus docs (or link to it).

As a part of this effort, I also made the example app work with our build
process and live alongside our Go codebase. This does mean that all modules
used by this service now live in our shared `go.mod` file; however, that
seemed like an acceptable tradeoff. If you know how to place the dependencies
into a unique `go.mod` file in the `pkg/metricsserverbackend` folder, please
go ahead. I just didn't know how to and it felt non-trivial.jje

Remaining work:
- Replace MySQL with Postgres
- Implement an endpoint that retrieves metrics
- Add a database table that stores the collected data
- Figure out deployment for this (Do we put it in CI? etc)


freeqaz:
Abandoning this in favor of #132

PR #175, opened 2021-10-23
One Lint File To Rule Them All

factoidforrest:
This is it, the final lint config.  I figured out how to merge all the lint files into one and have one single group of settings lint the whole project.  Surprisingly tricky but it works great now that I got it.  You can see from the changes just how much of our code was not getting linted before.

Caveats: Your IDE might be a bit unhappy and require manual configuration.  Open your settings, search `eslint`, and make it look like this and you should be set forever.
shot_008

This PR also contains the first parts of the Vue POC in a broken state but..doesnt really matter for now I don't think.  It was just important to get all these lint changes into master before they get stale and cause a lot of conflicts.

We should wait to merge this until everything else in flight merges if possible.  Then rerun it one more time and merge


factoidforrest:
@freeqaz I did do some ignores, especially in the metrics-backend-server where I straight up commented out some broken tests and so on.  Felt out of scope for me to debug it all since I barely know how that thing works :)

PR #196, opened 2021-11-04
use crypto/rand for randRead

cmiller01:
* randRead is needed to generate nonce and keys, which should be from crypto/rand
vs math/rand


go/util/crypto.go
```diff
@@ -101,7 +102,7 @@ func GenerateSaltsAndKey(token types.Token, secret string) types.SaltsAndKey {
 	hashable := sha3.Sum512([]byte(tokenStr))
 	seed := append([]byte(tokenStr), hashable[:]...)
 	seedInt := binary.BigEndian.Uint64(seed)
-	random := rand.New(rand.NewSource(int64(seedInt)))
+	random := mathrand.New(mathrand.NewSource(int64(seedInt)))
 
 	return types.SaltsAndKey{
 		Sp: GetRandomStringOfLength(keySize, random),
```
cmiller01, line 108: there's some interesting things going on here, maybe it's intended, but you're re-using the random reader on each `GetRandomStringOfLength` call below, so the ordering of these objects in the struct matters...maybe this is intended, but wanted to check

freeqaz:
Tests all passed, but the last line of the CI script fails to push to Slack. I'm going to land this and fix that issue in another branch. Thanks for the PR!

go/util/crypto.go
freeqaz: Yep, that's intended. It's how we deterministically derive these values.

freeqaz (approved):
LGTM!

PR #203, opened 2021-11-04
throw if same origin

factoidforrest:


freeqaz (approved):
LGTM -- this will be a nice UX to warn people about the security problems they face.

We may eventually need to make this ignorable... but for now this is fine.

PR #207, opened 2021-11-05
Fix broken links from previous bad config

freeqaz:
This is a basic change but it fixes a bad link that has been getting some traffic... Oi!


factoidforrest (approved):
nice hack haha

PR #217, opened 2021-11-11
Fix deployed tokenizer backend to work within the demo app

breadchris:
The tokenizer backend was not working when configured to be in the demo app due to some missing config values, bugged code, etc. 
- Update docs to include docker login command before deploying
- Refactor tokenizer backend auth provider config to not just be set by one environment variable 
- Turn on verbose logs when debugging production
- Add policy to tokenizer backend to push to cloudwatch metrics
- Deploy source map for js to the cloudfront bucket
- Fix bug which would incorrectly check origin of the frame


js/sdks/packages/cli/src/cdk-stack/stack.ts
```diff
@@ -105,9 +113,15 @@ export class LunaSecDeploymentStack extends cdk.Stack {
...
-    const sessionsTable = createDynamoDBTable(this, 'sessions-table');
-    const grantsTable = createDynamoDBTable(this, 'grants-table');
+    const keysTable = createDynamoDBTable(this, 'keys-table', {
+      retain: true,
+    });
+    const sessionsTable = createDynamoDBTable(this, 'sessions-table', {
+      ttl: '1h',
+    });
+    const grantsTable = createDynamoDBTable(this, 'grants-table', {
+      ttl: '1h',
```
factoidforrest, line 123: niiiice so dynamo throws old ones out now?  very good

factoidforrest: really fucking nice DB feature

factoidforrest: Actually I think we need this to be based on the "max grant time" config thing I wrote that takes in a configurable maximum grant time.  Otherwise the record could get deleted before the grant expired.  If you want to use the most naive approach I'd set it to 24h. 

Assuming that ttl configuration option uses the same time DSL as go's time.ParseDuration(string) I have a feeling we can just pass in whatever the config option was and forget about it.  

Heres that function that figured out the grant TTL
```golang
func (s *grantService) getGrantDuration(customDurationString string) (int64, error) {
	s.logger.Debug(
		"read custom grant duration from request:",
		zap.String("durationString", customDurationString))

	if customDurationString == "" {
		return time.Now().Add(s.grantDefaultDuration).Unix(), nil
	}

	customDuration, err := time.ParseDuration(customDurationString)
	if err != nil {
		return 0, errors.New("Grant duration parse failed, please use a supported duration format like 30m or 1h20m10s")
	}

	if customDuration > s.grantMaxDuration {
		return 0, errors.New("Grant duration set longer than configured maximum time")
	}
	return time.Now().Add(customDuration).Unix(), nil
}
```

factoidforrest: Sessions you might also want it to be a bit longer, since I'm assuming this is just for space/cost savings and ideally wouldn't colide with our systems logic.  Its nicer when the API responds intelligently that "session expired" or 
 something instead of just "i dont know what the fuck youre talking about" :)

js/sdks/packages/cli/src/config/types.ts
```diff
@@ -154,9 +169,14 @@ export interface DeploymentConfigOptions {
    * Only used when building locally. Specify the build location of where the built secure frame iframe exists.
    */
   frontEndAssetsFolder?: string;
+  /**
+   * Debug the deployed tokenizer backend. Defaults to false.
```
factoidforrest, line 173: which turns on verbose logging, I assume?

factoidforrest (approved):
Very nice.  Approving with the stipulation we think about those TTLs colliding with the users config

freeqaz (approved):
LGTM

js/sdks/packages/cli/src/cdk-stack/stack.ts
breadchris: yeah totally, I think sticking with the naive approach is reasonable for now. It isn't pressing to delete the entries in this table, but will help save people money. The sessions table only holds values that are used to prevent CSRF attacks so those are only really in use for like a second, so time wise those guys can come and go.

PR #224, opened 2021-11-17
Fix Docusaurus image by using MDX instead

freeqaz:
Docusaurus is very magical. I'm glad to learn more about MDX now! It's dope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 
 <p align="center">
-  <img src='docs/static/img/logo-black-text.svg' width='70%'>
+  <img src='docs/static/img/logo-black-text.svg' width='60%'>
 </p>
 
 ---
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,9 +53,30 @@ and also provide the foundation that LunaSec builds on to offer you additional s
 
 You can read more [here](https://www.lunasec.io/docs/pages/overview/features/) about what features LunaSec provides.
 
+## Live Demo
```
factoidforrest, line 56: We might want to say right here that

"If you just want to see lunasek running in a browser, click here to see the hosted demo.  If you want to run it locally..."

factoidforrest (approved):

README.md
freeqaz: I will when the Live Demo actually exists ;)

freeqaz:
I'm going to land this because the failing test seems to be erroneous.

PR #225, opened 2021-11-19
Service waiting without Docker

freeqaz:
This lets us safely use Docker-Compose Version 3+ without any issues now. And that's important for getting the demo app to work reliably on other people's machines.


breadchris (approved):
tested and reviewed this with Free as we try to bring up the live demo.

PR #226, opened 2021-11-19
Add the Public Demo config to the CLI

freeqaz:
There is still a bunch that's broken with this, but I'll walk you through it more tomorrow.

Basically:
- I'm using `docker context` via SSH to deploy the service using `docker-compose` directly.
- It's kind of jank. Paths have to be the same between your box and the Docker host, but that's doable.
- Volumes are passed currently in the `docker-config.yaml` file and they need to be removed for the live demo env.
- This error happens when it loads the secure frame because a host URI is bad still: `Blocked loading mixed active content “http://application-back-end:3001/.lunasec/secure-frame?state=c2af07f4-f6f4-4310-aa0b-ac7e9e92562a”`

I wrote the NGINX config and that seems to work properly. Overall, this seems pretty close now! We just need to sit down and debug the last remaining pieces. :)

### DIFF
--- a/js/sdks/packages/cli/src/docker-compose/lunasec-stack.ts
+++ b/js/sdks/packages/cli/src/docker-compose/lunasec-stack.ts
@@ -494,17 +494,17 @@ export class LunaSecStackDockerCompose {
     if (this.env === 'demo') {
       return 'localhost';
     }
-    if (this.env === 'tests' || this.env === 'hosted-live-demo') {
+    if (this.env === 'tests') {
       return 'application-back-end';
     }
     return undefined;
   }
 
   getLocalstackHostname() {
-    if (this.env === 'tests' || this.env === 'hosted-live-demo') {
-      return 'localstack';
+    if (this.env === 'demo') {
+      return 'localhost';
     }
-    return 'localhost';
+    return 'localstack';
   }
 
   getSecureFrameHostname() {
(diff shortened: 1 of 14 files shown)


js/sdks/packages/cli/src/docker-compose/lunasec-stack.ts
```diff
@@ -569,7 +632,10 @@ export class LunaSecStackDockerCompose {
 
     writeFileSync(composePath, dockerCompose);
 
-    const dockerEnv = this.getDockerEnv();
+    // TODO (freeqaz) Create a function for every environment that we support and
+    // have it merge in all of them using something like Ramda.pipe()
+    // As long as every function is referentially transparent, it's a clean abstraction.
+    const dockerEnv = generateNginxEnvConfig(this.env, this.getDockerEnv());
```
breadchris, line 638: this feels like a pretty confusing pattern since env values are going to be overwritten in a possibly confusing way. since we need to overhaul how the config is being created, I'm going to say this is ok for now since it works. creating CDK like constructs for each service will let us customize how each service presents itself to other services depending on what environment it is in:
```
tokenizerBackend = new TokenizerBackend(env);
applicationBackend = new ApplicationBackend(env, {
  tokenizerBackend,
});

class ApplicationBackend extends DockerComposeService {
  envName = 'APPLICATION_BACKEND';
  hostname = 'application-back-end';

  constructor(env, deps) {
    super(env, deps);
  }
}

class DockerComposeService {
  envVars: Record<string, string> = {}
  hostname: string | null = null;
  port: number | null = null;
  protocol: 'http' | 'https' | null = null;

  constructor(env, deps) {
    this.envVars = deps.reduce((dep, envVars) => {
      return {
        ...envVars,
        [dep
      }
    }, {});
  }

  getIntraNetworkName() {
    return `INTRA_NET_${this.envName}`;
  }

  getInterNetworkName() {
    return `INTER_NET_${this.envName}`;
  }

  getIntraNetworkValue() {
    if (this.hostname === null) {
      raise new Error('hostname is null');
    }
    ...
    return `${this.protocol}://${this.hostname}:${this.port}`;
  }

  getInterNetworkValue() {
    return `${this.protocol}://localhost:${this.port}`;
  }
}
```

breadchris (approved):
we will need to follow this up with more changes to the cli, but as long as this is working right now it is OK

PR #228, opened 2021-11-22
Mac OSX

factoidforrest:


factoidforrest:
@freeqaz Closing this for now.  Rather than trying to get things to build on Mac, we should just have a seperate nightly job that tries to start the existing containers in demo mode on mac, like how we have been testing manually.  That way it wont have to build them itself and we can kick the can with getting mac builds from source working.

PR #230, opened 2021-11-22
inject analytics into demo app

factoidforrest:
not 100% this is going to work/ not break things.  The analytics can be injected anywhere and this is a lazy solution.  untested pending fixes in master


factoidforrest:
we can do this a better less hairbrained way when we start building the demo app statically, so lets table this for now.  I think we could even use the same injector script from the docs

PR #232, opened 2021-11-24
Fixed simple tokenizer for local dev

factoidforrest:


freeqaz (approved):
This is only for local dev, fortunately. :)

PR #235, opened 2021-11-26
matrixed a no-lockfile build

factoidforrest:
people often blow the lockfile away as part of working and it would be nice to know ahead of time if the lockfile is all that is keeping things working, so that we can go and fix the dependency issue before its in master.


factoidforrest:
Yeah this appears to work perfectly, it caught the issue.  Its only 2 mins slower as well

breadchris (approved):
LGTM

PR #238, opened 2021-11-30
Security Blog Post

breadchris:
This blog post tells a story about what it is like to be a developer and care about security. LunaSec is proposed as a solution for people to use to solve their security issues with little effort on the developer.


freeqaz (approved):
LGTM

PR #239, opened 2021-11-30
Fix Documentation deployment job

freeqaz:
Small change. I didn't add this to the job, so the CI was failing. This should fix the problem though!


breadchris (approved):

freeqaz:
I'm just going to land this because it doesn't touch the tests at all.

PR #245, opened 2021-12-02
CSP blog post

factoidforrest:
Here's my blog post about CSPs.  My research definitely confirmed what you said @breadchris, they're useless for most sites.  Really interesting.  

I'm feeling good about most of the verbage here, mostly looking for a fact check and proof-read.


docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+// API callback
+alert('This is a malicious alert')({
+  // ...googles intended payload
+}
+);
+```
+
+Go ahead and try [the url from that script](<https://accounts.google.com/o/oauth2/revoke?callback=alert('This is a malicious alert')>) in your browser.
+You'll quickly see the problem.  Now an attacker that can embed HTML in your page has
+the ability to execute whatever javascript they want. Any un-sanitized field that leaks HTML into your page is now an immediate vulnerability.
```
breadchris, line 82: execute whatever javascript they want because they can use this jsonp endpoint maliciously, with a crafted payload, as the source of a script tag.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+
+There a great number of these vulnerable legacy endpoints, not just on googles domains but across the web.
```
breadchris, line 84: *Google's

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+
+Because this isn't a specific protocol but just a serverside (anti)pattern, there's no way for the browser
+to detect this behavior and disable it.  The only solution is to not allow scripts to load from any vulnerable domain, and 
+to host them yourself instead.
+
+ **The bigger your website, the harder this is to deal with.**
+
+If you have a small site and immediate control over the codebase, this could be an easy problem to patch.
+
+For large sites at big companies that load analytics payloads, third party auth, and other scripts from these vulnerable domains, it can be
```
breadchris, line 95: It might be nice to have an image of an example website shown with all the different 3rd party dependencies that get loaded.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+difficult to mitigate.  Heck, a lot of big companies have legacy JSONP endpoints on the _same domain as the site._  There's not much you can do about that.
+
+### File Upload Bypass
+Let's say your CSP is locked to only allow scripts from your domain.  
+
+If your site allows file uploads and downloads, including files that end in `.js`, a user can simply upload a file `malicious-code.js`,
+and load it as a script tag, directly from your domain. 
+
+```html
+<script src=www.yourdomain.com/myfiles/malicious-code.js></script>
```
breadchris, line 105: there is a caveat to this, the content type of the returned file needs to be `application/javascript` for this to work.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+These are just a few of the potential ways to bypass the CSP's protections.
+The list [goes on](https://book.hacktricks.xyz/pentesting-web/content-security-policy-csp-bypass).
+
+### Mitigations
+For large sites, there may be just too many chinks in the armor to treat the CSP as a functional line of defence, and many big sites have disabled it entirely. 
+[This paper](https://research.google/pubs/pub45542/) found that of those that use it **over 94% of sites have ineffective CSPs!**
+
+Long story short, there's a good chance your CSP isn't doing much to protect your site from XSS.
+
+Nonce-based CSP strategies can help mitigate the above attacks, where the server adds a random nonce attribute to every script tag and defines in the CSP.
```
breadchris, line 120: This is called "strict csp" https://csp.withgoogle.com/docs/strict-csp.html

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+
+The above problems with CSP are a great demonstration of why we've worked to try to isolate and secure a small portion of our user's pages, 
+instead of trying to completely secure the main sites themselves.  Big production web apps are just too large of an attack surface
+for the average company to fully audit for vulnerabilities.
+
+LunaSec and frameworks like it embed a tiny, stripped down, domain specific webpage that does one thing, totally focused on security.
+Because it's on a different domain, (in LunaSec's case, usually a subdomain you own) the main site can be vulnerable to the above issues without 
+exposing the sensitive data to XSS.  
+
+Separating out the security concerns into a separate domain is one of the few ways for large sites to employ a CSP that actually
```
breadchris, line 152: We could include what our csp policy looks like as an example and underline the fact that this policy will only ever look like that because it is a locked down policy. CSP becomes problematic/useless when it needs to change as frequently as you ship features in your application.

breadchris (changes requested):

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+I'm a normal, non-security focused developer, but I work for a security company.  I wasn't surprised to learn that our 
+services use a good, well-thought-out CSP to protect our users. 
+
+What did come as a *big* surprise to me is that most companies, even huge, well thought of firms, have CSPs that don't
+protect from *anything*, or don't even use them at all. For the most part, it's not their fault and there's **not much they can do
+about it.**  More on that below.
+
+### How CSPs work
+
+The Content Security Policy (CSP) is a response header or `<meta>` tag set by the server when a webpage first loads.  It controls
```
freeqaz, line 35: You should link out to this page to define CSP.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+protect from *anything*, or don't even use them at all. For the most part, it's not their fault and there's **not much they can do
+about it.**  More on that below.
+
+### How CSPs work
+
+The Content Security Policy (CSP) is a response header or `<meta>` tag set by the server when a webpage first loads.  It controls 
+what the page is allowed to load and, most importantly, from _where_.  The idea is to act as a last-line-of-defense against
+ loading something malicious onto your page.  
+
+If you've never seen one before, they're actually pretty simple.
```
freeqaz, line 39: Maybe throw the `<truncate>` tag here?

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+That CSP also means that inline scripts like `<script> alert('Hi I'm a script') <script>` won't be executed, which is great for security.
+
+So, if we trust our server and Google's server not to do something malicious, we would expect to be safe.  
+
+Unfortunately, this CSP is still totally insecure, because:
+
+## CSP is swiss cheese
+Studies have found that the vast majority of sites using CSPs are still vulnerable to script injection.  Here are some of the most common attacks.
+### JSONP Bypass
+JSONP is a hacky way to load cross domain javascript in the browser from the mid 2000's, that has since been replaced by CORS.  More of a pattern than a protocol,
```
freeqaz, line 64: Extraneous comma here (the second half after the comma is a continuation and not a dependent clause).

Also, I substituted "replaced" with "superceded" because I believe that's clearer. They both exist in parallel, but JSONP is legacy.

Finally, I replaced `javascript` with `JavaScript` because that feels a little better. I'd also accept `JS` too.

https://www.scribbr.com/language-rules/commas/

Options:

**Rewrite second half to be a dependent clause**:

> JSONP is a hacky way to load cross domain JavaScript in the browser from the mid 2000's, and it has since been superceded by CORS.

**Remove comma to make it one long sentence**:

> JSONP is a hacky way to load cross domain JavaScript in the browser from the mid 2000's that has since been superceded by CORS.

**Make the middle a dependent clause**:
> JSONP is a hacky way to load cross domain JavaScript in the browser, added in the mid 2000's, that has since been superceded by CORS.

That's my thought process on this. Small tweaks like this will make the text flow a lot more nicely, imo.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+Let's say your CSP is locked to only allow scripts from your domain.  
+
+If your site allows file uploads and downloads, including files that end in `.js`, a user can simply upload a file `malicious-code.js`,
+and load it as a script tag, directly from your domain. 
+
+```html
+<script src=www.yourdomain.com/myfiles/malicious-code.js></script>
+```
+
+You have to be very careful about what your users can upload.
```
freeqaz, line 108: This is further complicated by browsers attempting to parse arbitrary documents into JS or HTML, also. You should include a reference to `nosniff` in this section.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+You have to be very careful about what your users can upload.  
+
+### Other vulnerabilities
+These are just a few of the potential ways to bypass the CSP's protections.
+The list [goes on](https://book.hacktricks.xyz/pentesting-web/content-security-policy-csp-bypass).
+
+### Mitigations
+For large sites, there may be just too many chinks in the armor to treat the CSP as a functional line of defence, and many big sites have disabled it entirely.
```
freeqaz, line 115: I'd probably avoid the use of the word "chinks" here. While the technical definition of it is fine, the slang definition is most definitely _not fine_.

https://www.urbandictionary.com/define.php?term=Chinks

Probably "dents", "gaps", or "scratches" would be better.

freeqaz: Also, "defense" is probably the better spelling here (because the rest of our documents are spelled with American English, lol).

https://www.grammarly.com/blog/defence-defense/

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+### Mitigations
+For large sites, there may be just too many chinks in the armor to treat the CSP as a functional line of defence, and many big sites have disabled it entirely. 
+[This paper](https://research.google/pubs/pub45542/) found that of those that use it **over 94% of sites have ineffective CSPs!**
+
+Long story short, there's a good chance your CSP isn't doing much to protect your site from XSS.
+
+Nonce-based CSP strategies can help mitigate the above attacks, where the server adds a random nonce attribute to every script tag and defines in the CSP.
+Embedding a script tag smart enough to clone the nonce isn't possible without javascript.  
+
+Unfortunately this strategy is the most work to implement and is rarely used.
```
freeqaz, line 123: I'd probably mention that the increasingly popular JAM stack pattern of serving documents off of a CDN makes this difficult because the Nonce must be random for it to be secure.

...

There really aren't very many resource available online to talk about this! Only this Tweet[0]. It probably warrants a separate blog post where we can do a deep dive into "How to make your Jamstack app secure with CSP". In order to _actually_ do this, you have to use something like `Lambda@Edge` or CloudFront Workers to rewrite the response (see this[1] post).

Basically, choose either a Nonce or a low-latency CDN. (Or the middle ground of writing code to rewrite the cached response)

0: https://twitter.com/mikesherov/status/1378373386171191299

1: https://medium.com/@csarkosh/my-experience-getting-an-a-from-mozillas-observatory-tool-on-aws-f0abf12811a1

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
...
+
+```html title="Script Tag"
+<script nonce="randomString" ...
+```
+
+Defining a nonce and a domain `script-src` together means that _either_ the domain or the nonce can be used, which seems like a
+gotcha that could catch many users.
+
+## How LunaSec and other tools use CSP effectively
+LunaSec is an Open Source security product we wrote, designed to help unhardened web-apps handle sensitive data. The main
```
freeqaz, line 137: Rewrite to, "[...] designed to harden web apps that handle sensitive data against leaking sensitive data, even _when_ CSP fails to protect you from XSS."

The word "unharden" is... the opposite of what we should be saying!

freeqaz: also, `s/web-app/web app/` because that's how we have it everywhere else in our docs.

freeqaz (approved):
Mostly nits, but please make these changes before landing!

Otherwise, this looks great :)

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,153 @@
+---
+title: Don't trust your Content Security Policy
```
freeqaz, line 2: We should probably wordsmith this title a bit too. It'd be good to make it something that will rank well for SEO.

Generally, if we want somebody to find this while they're coding, then you'd want to use words that are closer to a query you might expect.

"Understanding the security implications of adding a Content Security Policy"

Something like that. We should talk in the chat and go through some options before we post this up. This is the #1 metric that will tell us if this post will get search traffic or not.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,208 @@
...
+
+Here is an example CSP header that loads with one of LunaSec's iFrames: 
+```text
+Content-Security-Policy: 
+style-src 'unsafe-inline' 'nonce-r54BDFpZ8Nkb9IffY7JwJA';
+script-src 'nonce-r54BDFpZ8Nkb9IffY7JwJA';
+object-src 'none';
+frame-ancestors https://your-site.com;
+base-uri 'none';
+report-uri http://localhost:5004;
```
breadchris, line 196: we probably want to change this to `http://lunasec-secure-backend.your-site.com/cspreport`. `http://localhost:5004` is just a placeholder for now.

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,208 @@
...
+```text title=CSP
+script-src 'nonce-randomString';
+```
+
+```html title="Script Tag"
+<script nonce="randomString" ...
+```
+
+Defining a nonce and a domain `script-src` together means that _either_ the domain or the nonce can be used, which seems
+like a gotcha that could catch many users.
```
breadchris, line 168: I would throw this in here https://csp-evaluator.withgoogle.com/ as a call to action: "see for yourself how secure your site's CSP policy is"

docs/blog/2021-12-01-csp.md
```diff
@@ -0,0 +1,208 @@
...
+list [goes on](https://book.hacktricks.xyz/pentesting-web/content-security-policy-csp-bypass).
+
+### Mitigations
+
+For large sites, there may be just too many gaps in the armor to treat the CSP as a functional line of defense, and
+many big sites have disabled it entirely.
+[This paper](https://research.google/pubs/pub45542/) found that of those that use it **over 94% of sites have
+ineffective CSPs!**
+
+Long story short, there's a good chance your CSP isn't doing much to protect your site from XSS.
```
breadchris, line 151: "The only effective way to enforce CSP is by abiding by what is known as 'strict csp'" https://csp.withgoogle.com/docs/strict-csp.html

docs/blog/2021-12-01-csp.md
breadchris: Also you could add "For an idea how effective CSP is, https://google.com has no CSP headers set. https://mail.google.com is a prime example of what a strict csp policy looks like, it simply uses a nonce to get all the benefits of a strong CSP policy."

breadchris (changes requested):
Some additional detail that I think would be nice to link people and a change to our CSP policy.

PR #246, opened 2021-12-02
Replace yargs with cmd-ts

breadchris:
Yargs just plain doesn't work when you have multiple flags in a command (ie. --flag1 --flag2) as the second and subsequent flags will all report "false" even though they should be reporting true. cmd-ts doesn't have this problem and is overall a better library to use as it is fully written in typescript.


factoidforrest (approved):
neat LGTM

PR #248, opened 2021-12-02
Speed up CI by migrating to Yarn 3

freeqaz:
This requires committing all of our dependencies which definitely balloons the size of this PR. You might need to pull this down manually in order to effectively review it.

In this change, I've done a few things:
- Replaced Yarn 1 with Yarn 3
- Switched to [PNP](https://yarnpkg.com/features/zero-installs) mode
- Fixed some bugs in the Docker Env CLI refactor
- Switched to Alpine for the base image from Ubuntu
- Removed the `precached-dependencies` Docker image dependency
- Fixed a ton of bugs in the `package.json` files where we relied on packages implicitly (Yarn 3 caught these)
- Removed dependency on AWS CLI in the application-back-end to get the S3 Bucket name

There is a probably more here too... but I'll leave it here.


factoidforrest:
We can probably drop the merged/not merged matrixing from everything except the integration test as well. 

I'll get the no-lockfile job fixed soon, I need to go read yarns source and figure out how it is checking for "CI" mode so we can hack around them locking us into it.

factoidforrest:
🎵 it builds 🎵

PR #252, opened 2021-12-08
Split CI job up into parallel workflows

freeqaz:
This should help us keep everything bug-free and fast.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -14,7 +14,7 @@
 #
 # This is a basic workflow to help you get started with Actions
 
-name: CI
+name: Integration Tests
 
 # Controls when the workflow will run
 on:
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -33,11 +33,6 @@ on:
 jobs:
   integration-test:
     runs-on: ubuntu-latest
-    env:
-      # avoid warnings like "tput: No value for $TERM and no -T specified"
-      TERM: xterm
-      DOCKER_BUILDKIT: 1
-      COMPOSE_DOCKER_CLI_BUILD: 1
     strategy:
       matrix:
         merge: ["merged", "not-merged"]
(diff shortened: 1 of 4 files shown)


factoidforrest:
Really nice use of the composite steps.  I definitely don't mind getting rid of all that skip logic that wasn't really doing anything, and for the unit tests and linting we really don't need to run in more than one environment.  Nice changes!

factoidforrest (approved):

.github/workflows/composite/setup-docker-ci/action.yml
```diff
@@ -0,0 +1,54 @@
...
+#      if: ${{ inputs.skip_docker != 'false' }}
+
+    # specifically tag this bootstrap container to prevent rebuilds
+    - name: Build Lerna Bootstrap container
+      shell: bash
+      env:
+        # avoid warnings like "tput: No value for $TERM and no -T specified"
+        TERM: xterm
+        DOCKER_BUILDKIT: 1
+        COMPOSE_DOCKER_CLI_BUILD: 1
```
factoidforrest, line 42: Can we not put these at the top anymore because of composite mode?

.github/workflows/integration.yaml
```diff
@@ -86,9 +61,14 @@ jobs:
       if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled != 'false' }}
       timeout-minutes: 15
 
-    - name: Delete and Regenerate Lockfile
-      if: ${{ matrix.lockfile == 'no-lockfile' }}
-      run: rm yarn.lock && CI="" yarn install #currently broken
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '14'
```
factoidforrest, line 66: 16 is LTS now

factoidforrest: And we can probably turn that cache off, its really cool and really fast but technically we aren't using it for anything

.github/workflows/lint-code.yaml
```diff
@@ -0,0 +1,37 @@
...
+      uses: ./.github/workflows/composite/setup-docker-ci
+      with:
+        merge_master: ${{ matrix.merge == 'merged' }}
+        skip_docker: 'false'
+        event_name: ${{ github.event_name }}
+
+    - name: dependencies
+      env:
+        IS_LUNASEC_CI: true
+      run: yarn install --immutable --immutable-cache --inline-builds
```
factoidforrest, line 29: Why are we doing this bare-metal install?  This should be happening in the docker container, called in the above setup step right?  If this is the security "check" we talked about before, maybe we should just put that in its own step since it has nothing to do with any of the rest of this stuff and doesn't depend on docker.

.github/workflows/integration.yaml
```diff
@@ -168,15 +134,12 @@ jobs:
...
-      #TODO: there may be a more idiomatic way to do this
-      - name: Check if we should skip this build
-        id: job-canceller
-        run: echo "::set-output name=cancelled::${{ matrix.merge == 'merged' && github.event_name != 'pull_request'}}"
-
-      # merge with master(or whatever target branch) so we are actually testing what will happen after PR merges, not just this branch
-      - name: Merge target branch
-        run: git merge origin/${{ github.event.pull_request.base.ref }}
-        if: ${{ matrix.merge == 'merged' && github.event_name == 'pull_request' }}
+      - name: Setup Environment
```
factoidforrest, line 137: Maybe it is the diff being weird, but didn't we do this up above, and isnt this the end of the job now?

factoidforrest (approved):

.github/workflows/composite/setup-docker-ci/action.yml
```diff
@@ -0,0 +1,54 @@
...
+  steps:
+    - name: Set up Docker BuildKit
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+      with:
+        install: true # sets buildx as the default for docker, which should apply to docker-compose commands
+        driver: docker
+#      if: ${{ inputs.skip_docker != 'false' }}
+
+    - uses: actions/setup-node@v2
```
factoidforrest, line 29: Actually do we need node at all anymore?  Isnt it all in docker?

.github/workflows/composite/setup-docker-ci/action.yml
freeqaz: I'll test this to verify.

PR #260, opened 2021-12-10
Avoid using user input as format string

Maxdamantus:
When I first saw this post, I thought the issue was limited to incorrect usage of the `Logger` API. Hopefully this change will make it clearer to others who might also think this.

The first string argument to a logging method is meant to be a format string, which is normally a constant string rather than something derived from user input.

The JNDI lookup still occurs when the user input is not included in the format string, so this change should make it clear that it's not an issue with misplacement of user input.


Maxdamantus:
Sorry, I'm closing this because I realised the particular code sample uses log4j 1 rather than log4j 2. The issue seems to exist in both versions of log4j, though the original code used here seems to be the correct way to log in version 1.

PR #262, opened 2021-12-10
State that CVE has been allocated

chris-x86-64:
Looks like there's now a CVE. Source https://github.com/advisories/GHSA-jfh8-c2jp-5v3q


freeqaz:
Thanks for adding this! You beat me to the punch while I was making a PR for this. I added this in #263 just a few minutes after you!

PR #269, opened 2021-12-10
log4j download page

pjfanning:
Apache projects' primary approach to releases is to provide downloads via the apache project web sites - not via github.


factoidforrest (approved):

factoidforrest:
Thank you!

PR #270, opened 2021-12-10
Update 2021-12-09-log4j-zero-day.md

nosnilmot:
Update example vulnerable code to use paramaterized logging instead of string concatenation.

This should help avoid people assuming that the problem is the string concatenation itself (which has previously been a source of problems).


freeqaz (approved):
Thanks for this! I'm going to land this onto the branch I'm working on, and then ship this in a minute.

PR #276, opened 2021-12-10
Correct iPhone statements

timbuchwaldt:
Given that iPhones for sure don't run Java and the network is operated by Apple, we can safely assume it's evaluated in their backend code.


factoidforrest:
I agree that that part of the post was confusing.  We've fixed it upstream.

PR #277, opened 2021-12-10
Fix example to use log4j2

nosnilmot:
pedantry: make the example code actually valid log4j 2 use, not log4j 1.x

PR #260 was abandoned because the submitter recognized the example was using log4j 1.x syntax, but if that PR hadn't been abandoned it would have saved me time checking that paramaterized usage was vulnerable and submitting PR #270 to make the same change (and I'm not sure 1.x is vulnerable anyway)


freeqaz:
Thanks for catching this! I'll land this now :)

freeqaz (approved):

PR #279, opened 2021-12-10
make demo backend use pm2

factoidforrest:
I'm not sure why I wasn't able to reproduce this crash for so long but now I am. Maybe node 16...not sure. 

Anyway, skipping ts-node-dev and running the built JS directly seems to both keep the node process from crashing, and restart it gracefully if it did happen to crash.  

One caveat to keep in mind is that pm2 doesn't pass in locally declared variables like 
`VARIABLE=xyz pm2 start i-cant-read-the-variable.js` but if you use export it works fine like `export VARIABLE=123` So I believe it will properly read things out of docker env, no trouble.


factoidforrest:
merged into the monster PR

PR #283, opened 2021-12-12
Add log4shell scripts + Mitigation blog post

freeqaz:
This should help people out.


factoidforrest:
nice.  I was wondering if we might want to make this part of our main CLI.  For business reasons.

docs/blog/2021-12-11-log4j-zero-day-migitation-guide.md
```diff
@@ -0,0 +1,173 @@
...
+:::
+
+<!--truncate-->
+
+## 1. Determine if you're impacted
+
+This vulnerability affects anybody who's using the log4j packages (`log4j-core`, `log4j-api`, etc). That means it's
+primarily Java, but other languages like Scala, Groovy, or Clojure are also impacted.
+
+### By Checking Package Hashes
```
factoidforrest, line 49: Say up top that we wrote a CLI tool to make it easy.

factoidforrest: Yeah i think we should move all of this into option 2

js/sdks/packages/log4shell/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
factoidforrest, line 1: There's no JS in this tool whatsoever, and its not really one of our SDKs.  We can publish packages out of the `/tools` directory (nothing stopping us from doing that).  That would make a bit more sense to me, personally.

PR #284, opened 2021-12-13
improve DNS test paragraph

factoidforrest:
I didn't end up linking to the source as that tool is kinda big, and you can find the source on the website itself really easily


factoidforrest:
Okay ready to merge IMO, please land when ready

freeqaz (approved):

PR #296, opened 2021-12-14
Adding command for running log4shell hotpatch server

breadchris:
The log4shell hotpatch server can be used to exploit the log4shell vulnerability to prevent further exploitation until a more permanent mitigation is rolled out.


factoidforrest:
We should tag a new version yeah?  In the constants/version.  1.1.0?

PR #297, opened 2021-12-14
Mitigation edits forrest

factoidforrest:
Sorry for the doubled up commits, shouldn't have reused the branch. 

Adds the mailchimp contact form.  Hitting tab to go to the next input doesn't work for me, docusaurus is not a fan of forms.  Truly.   This lib doesn't let you modify tabIndex.  For now I think we just accept it.  

A few other rewordings here and there, including a factual correction of not mention log4j-api because its not actually vulnerable according to people yelling on reddit at us.

I took out some of the mentions of "secure by default" on the resources list because they didnt have much to do with what we were listing, like scanners and mitigation tools.


freeqaz (approved):
This is great, thank you!

PR #298, opened 2021-12-14
Adds information about Thread Context Map Property Substitutions that might still be vulnerable even with formatMsgNoLookups=true

kmindi:
Adds information about Thread Context Map Property Substitutions that might still be vulnerable even with formatMsgNoLookups=true

**Please verify before just merging** I could just create a minimal example and minimal checks if 2.14.1 is really still vulnerable when using ctx. Great to hear if someone could confirm this.

I'm not sure about the other property substitutions and in general how popular these are, at elast for some commercial products ctx seems to be used.

### DIFF
--- a/docs/blog/2021-12-12-log4j-zero-day-mitigation-guide.md
+++ b/docs/blog/2021-12-12-log4j-zero-day-mitigation-guide.md
@@ -261,6 +261,12 @@ Or you can set this using the JVM arguments environment variable.
 
 `JAVA_OPTS=-Dlog4j2.formatMsgNoLookups=true`
 
+#### Does not fix Thread Context Map Substitutions
+
+Apparently this Option's mitigations do not prevent user controlled context that is passed via ThreadContextMap/ctx (and possibly other [PropertySubstitutions](https://logging.apache.org/log4j/2.x/manual/configuration.html#PropertySubstitution)). See demonstration here https://github.com/kmindi/log4shell-vulnerable-app 
+
+E.g. if your pattern layout contains something like `${ctx:apiversion}` the exploit might still work, at least a DNS lookup is performed.
+
 ### Option 3: JNDI patch
 It's possible to [modify the JNDI in place](https://news.ycombinator.com/item?id=29507263) to stop the attack at the language level.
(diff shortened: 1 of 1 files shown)


kmindi:
Got at least one confirmation https://github.com/apache/logging-log4j2/pull/608#issuecomment-993469509

SpComb:
Note in particular that the thread context lookups seem to also be affected under 2.15.0. AFAIK the 2.10.x mitigation mechanism and 2.15.0 fixes to disable *message* lookups are very similar, so the same issue affects both.

I believe the JNDI restrictions in 2.15.0 / disabling in 2.15.1 will still stop the RCE and DNS infoleaks, but the context lookups seem to get expanded recursively.

pl-semiotics:
I can also confirm this issue exists and affects the formatMsgNoLookups=true case and 2.15.0 (which does not have the jndi restrictions.) My example is [here](https://goto.ucsd.edu/~lucy/ephemera/d1a2769b2c967dca4db6cb721e109cb0e73c263e8c2dd03e039d8426f950631f). I have reported on the log4j JIRA as well as [LOG4J2-3221](https://issues.apache.org/jira/browse/LOG4J2-3221).

freeqaz:
Howdy y'all, I'm just waking up and looking at this. Give me a bit to dig through this and verify what's going on before I push this out. Thanks for the context!

SpComb:
Just in: https://lists.apache.org/thread/83y7dx5xvn3h5290q1twn16tltolv88f

factoidforrest:
Wow so that would mean that 2.15.0 is also vulnerable?  Yeah.  Wow.  I'll start making the changes to the blog, this affects a lot of stuff.

rgmz:
@factoidforrest the impact to 2.15.0 is limited because it can only connect to whitelisted IPs (localhost), in addition to the exploit criteria being more specific. The main concern, imo, is applications using < 2.15.0 that have applied the noLookup flag as a mitigation.

factoidforrest:
Yeah, we just reproduced the RCE and DOS on 2.14 and 2.15 respectively.  New blog post about this in the works

quaff:
Please add notes that `%X{}` `%mdc{}` `%MDC{}` are not affected, only `${ctx:}` is affected.

PR #302, opened 2021-12-14
Fix renamed directory in doc

natrem:


freeqaz (approved):
Nice catch -- thank you!

PR #305, opened 2021-12-15
Scanner finds 2.15

factoidforrest:
reports them with lower severity level.  This adds a feature but should break anything, so a minor version change.  Also you can put flags AFTER the scan command now, they dont have to be before

### DIFF
--- a/tools/log4shell/README.md
+++ b/tools/log4shell/README.md
@@ -24,6 +24,7 @@ docker run --network=host log4shell
 
 or 
 
+Make sure you have Maven installed, then:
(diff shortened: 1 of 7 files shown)


factoidforrest:
#298

freeqaz (approved):

freeqaz:
Maybe we should log out the CVE for each severity, also? I feel like that would be the most clear for people trying to get guidance around this.

freeqaz:
I went through and made that change. I think it's clearer for people.

PR #308, opened 2021-12-15
Add jar patcher command to log4shell cli

breadchris:
The jar patcher will attempt to remove the JNDILookup class from log4j dependencies that have been discovered.

Read more about this approach here: https://www.lunasec.io/docs/blog/log4j-zero-day-mitigation-guide/#option-3-jndi-patch

We can possibly use this tool to perform the fix as well: https://github.com/nccgroup/log4j-jndi-be-gone

https://www.trendmicro.com/en_us/research/21/l/patch-now-apache-log4j-vulnerability-called-log4shell-being-acti.html
> For 2.0-beta9 to 2.10.0, remove JndiLookup.class from class path: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class.


mikebell90:
Please dont do bytte buddy or dynamic patching. There are many many flaws to this technique and IMO it's mostly showing off and not realistic.

breadchris:
@mikebell90 i completely agree, simple is def better here. My approach for implementing this will be to only remove files that match known hashes like I have when scanning so that the risk of a false positive is very low.

Issue #301, opened 2021-12-14
No detection of log4j repackaged inside shaded jar


natrem:
The efficiency of checksum comparison has one downside: it does not find tools that repackage log4j.

For instance Elastic java agent version 1.28.0 is vulnerable (See [announcement here](https://discuss.elastic.co/t/apache-log4j2-remote-code-execution-rce-vulnerability-cve-2021-44228-esa-2021-31/291476)) but not detected.
From what I found out the [1.28.0 agent](https://github.com/elastic/apm-agent-java/releases/v1.28.0/) repackages log4j and changes the packages which results in a new class.

Do you think this class and similar ones should be added to the checksum list? IMO it serves the purpose of the tool (great job BTW)

natrem:
Note that a possible improvement would be to add an external hash table file to add custom checks;
I started modifying the tool for some usage and I can dig this topic if you're open to it.

factoidforrest:
Ah, great catch.  We were worried about anything that would change the hash.  If you have aleady started modifying the tool, we would love a PR that improves it.  It would be great if it could read a custom file like  --custom-hash-file file

rgmz:
I've checked with a few vendor SCA tools and they all get tripped up by this. I'm not sure if there's a foolproof method to scan repackaged/uber jars beyond searching for the presence of the JndiLookup file.

factoidforrest:
We were thinking it might be possible to use Java's class and method parsers to find vulnerable Log4j.  It would be a more heuristic approach to detecting the issue and it could come with its own problems. 

For now, if you discover any more vulnerable hashes in common packages(like the above apm-agent-java) please PR them in.

natrem:
With jar shadowing - as is the case with Elastic APM - even the package is changed so you must consider all classes named *JndiLookup* (and a few others present in your hash table).
I'll propose the APM addition when I get all impacted versions.

PR #319, opened 2021-12-15
Scan library before browsing it and detect Elastic APM

natrem:
- add Elastic APM hashes
- scan the cksum256 of an archive before diving into it. 
 
Spotting a known archive avoids reading its content. In the case of APM the lib is usually present on the server rather than embedded within the binary (such as a spring-boot app).

This relies on the existing function but may not be placed where you'd prefer it.

I perform local test on all versions but did not impact the ``test.sh`` script

This addresses my usage for #301


freeqaz:
Thanks for the PR @natrem! I'm way too tired to review this right now (or fix the tests that are failing because of missing secrets), but we'll be back tomorrow to take a look at this. See you on the other side :)

freeqaz (approved):
Alright, I have returned. This looks good to me, and I'll go ahead with landing this now!

factoidforrest:
We will have a release going out in the next couple hours probably

PR #333, opened 2021-12-16
Basic technical analysis of the Log4Shell exploit

freeqaz:
I could probably write more and add to this, but I'll leave it at this stage for now so that we can discuss what to add.


freeqaz:
I'd still like to add a bit at the end about the decompiler but I'll do that tomorrow. For now, I'm going to land this because it's already pretty good!

PR #334, opened 2021-12-16
add ear extensions to ease scanning

acollign:
This PR adds the EAR[0] extension to the scanner.

[0] EAR (Enterprise Application aRchive), this is used in the JEE world for deployment

### DIFF
--- a/tools/log4shell/constants/fs.go
+++ b/tools/log4shell/constants/fs.go
@@ -18,5 +18,6 @@ const (
 	JarFileExt   = ".jar"
 	WarFileExt   = ".war"
 	ZipFileExt   = ".zip"
+	EarFileExt   = ".ear"
 	ClassFileExt = ".class"
 )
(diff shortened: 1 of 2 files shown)


freeqaz (approved):
Thanks for this! I've learned so much about Java because of this exploit. JARs have EARs... it's terrifying

acollign:
Hi @freeqaz ,

> Thanks for this! 
You're welcome! I am glad this tiny contribution is useful.

> JARs have EARs... 
Well it is the opposite. EAR[0] files bundle JAR[1], WAR[2] and other files. Since it is quite frequent to use the zip archive format for distributing softwares, you may have a ZIP with multiple EARs containing JARs and whatever you may think of!

That is why all these formats need to be supported as "entry" file and nested file.

> it's terrifying
Yeah, it is ;-). At the end, EAR[0], JAR[1] and WAR[2] are ZIP archives following different conventions. I cannot say whether that is more or less terrifying, though!

[0] https://en.wikipedia.org/wiki/EAR_(file_format)
[1] https://en.wikipedia.org/wiki/JAR_(file_format)
[2] https://en.wikipedia.org/wiki/WAR_(file_format)

PR #362, opened 2021-12-18
Update 2021-12-09-log4j-zero-day.mdx

yburkouski:
preventing the situation when someone will try to override non changed class from a newer version and have the wrong conclusion that the issue is fixed


factoidforrest:
I see what you're trying to say.  Maybe we can clarify this and get it in.

breadchris:
I am not sure I understand the change in wording here. From my understanding, removing the `JndiLookup.class` file will prevent the `jndi:` format directive from being able to be resolved (since the code to perform this action just doesn't exist). 

Yes you are correct that the changes made to the log4j codebase for patching these security vulnerabilities are _not_ changing any code in `JndiLookup.class`, they have been primarily to the `JndiManager` class, removing the `JndiLookup.class` file from the log4j library will always patch this issue because it prevents the jndi string format directive from ever functioning fully.

Issue #361, opened 2021-12-18
2.16 vulnerable, expect 2.17


OlivierMasit:
Hi it seems that news vulns have been found on 2.16
Not clear yet from what I read...

https://issues.apache.org/jira/plugins/servlet/mobile#issue/LOG4J2-3230

OlivierMasit:
Confirmed https://nvd.nist.gov/vuln/detail/CVE-2021-45105

breadchris:
@OlivierMasit thanks for the heads up! We have also seen this and will probably include this as an update to an existing post, but since the severity is so low, I don’t know if we will make a dedicated blog post for it.

freeqaz:
I'm closing this out now that we've updated the guidance (it's mid-deploy now). We should probably write up a post dedicated to 2.17.0 and the DOS issue, but I might not have time for that until tomorrow. Too many IRL tasks piling up!

PR #367, opened 2021-12-19
update guidance to use 2.17.0

breadchris:
#361 

We have seen yet another vulnerability reported and so we are updating our guidance to reflect this.


freeqaz (approved):
I made some changes to update the dates everywhere with this new advice. I'm going to land this now.

PR #370, opened 2021-12-20
Updated Log4Shell logo

anomal:
There is a seashell made of a scroll (an ancient medium of logging) in the shape of a 'J' to symbolize Java and the 'j' in log4j. The `>` symbolizes the shell prompt that can be accessed by the attacker. The `_` as the cephalopod's pupil symbolizes the shell/terminal cursor. The cephalopod's tentacles spell out `${$$}`, which represents one form of the malicious input string.

This image is available on the public domain. https://commons.wikimedia.org/wiki/File:Log4Shell_cephalopod.png

### DIFF
--- a/docs/blog/2021-12-09-log4j-zero-day.mdx
+++ b/docs/blog/2021-12-09-log4j-zero-day.mdx
@@ -2,7 +2,7 @@
 title: "Log4Shell: RCE 0-day exploit found in log4j 2, a popular Java logging package"
 description: Given how ubiquitous log4j is, the impact of this vulnerability is quite severe. Learn how to patch it, why it's bad, and more in this post.
 slug: log4j-zero-day
-image: https://www.lunasec.io/docs/img/log4shell-logo.png
+image: https://upload.wikimedia.org/wikipedia/commons/a/a6/Log4Shell_cephalopod.png
 date: 2021-12-19
 keywords: [log4shell, log4j, log4j2, rce, java, zero-day]
 authors:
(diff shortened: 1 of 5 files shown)


breadchris:
wow this is very impressive @anomal! The symbolism is really quite striking and it is obvious you have put a lot of time and effort coming up with this :) 

I am not sure that updating all of our existing posts to include this logo is the best idea given that people have come to know the current logo to be associated with the vulnerability. 

That said, I really really dig this logo and would like to include this somewhere. Given how technical the symbolism is in this logo, how would you feel about us including this on our CLI tool's page: https://github.com/lunasec-io/lunasec/tree/master/tools/log4shell, with the description of the logo, as I think it will be pretty well received over there.

Thank you again for proposing this change :)

anomal:
Sure, you can include it anywhere. It is on the public domain now.

I hoped that there would be a better official logo than the MS Paint style one, to convey the seriousness of the vulnerability to middle management types.  I think Heartbleed seemed scarier to non-technical people, because of the name and logo, but Log4Shell is arguably worse.

Recently, I thought of a better version of the same idea in the colours and style of the Java logo  However, I'm just a Java developer who learned how to use Inkscape over the weekend. I hope a professional designer can make a more professional version of the same idea.

breadchris:
LOL @anomal you bring up a really good point! Unfortunately at this point I think it will be difficult to change what logo people will use to reference the vulnerability.

I personally think there is some validity to having the logo put together the way that it has since there was an initial mad scramble to understand what the heck this is and I think the logo kind of reflects that lol. I think if we were going to make updates to the main logo, we would update it to be more cleaned up and official looking so that someone could include it in a presentation.

Thank you for spending the time to learn inkscape to put this together!

PR #382, opened 2021-12-21
Build log4shell w/out linking to MUSL C library

hartzell:
Build log4shell w/ CGO_ENABLED=0 enabled so that the resulting binary isn't linked to Alpine's MUSL C library, resulting in a binary that runs on CentOS.

I won't be able to sign the CDA, so this is another "convene a tiger team and reimplement this change" kind of thing.


breadchris:
@hartzell thanks for another one! I will push a patch here soon that addresses the two PRs you have made. I'll close them out once I address them.

Thanks for the help!

breadchris:
see #386

PR #388, opened 2021-12-26
Blog post - Working backwards from log4shell to see why we built lunasec

factoidforrest:
WIP but ready for first review.  May need more references to Log4shell.  This blog post walks the line between interesting story, advertisement, and documentation, and I'm not sure if it achieves all three in a clean way.

The focus on anything too technical here is pretty cursory.  I think, in general, its enough to explain here that the product can protect you from RCE without getting into the implementation details of exactly how, such as auth-proxy strategies, "security levels", or multi-key encryption algorithms


factoidforrest:
Nice Chris, let me do one more typo pass on that and maybe we can ship it..phew haha

factoidforrest:
@freeqaz ready for merge or any changes you want to make

PR #409, opened 2022-01-08
Scan running processes for linux

breadchris:
For containers and other environments where just scanning running processes is sufficient, we can now do this for at least Linux. 

Supporting Windows and MacOS is a little more challenging since it isnt as simple as processing the procfs :P


factoidforrest:
Wow this is amazing

tools/log4shell/main.go
```diff
@@ -130,6 +130,10 @@ func main() {
 						Name:  "archives",
 						Usage: "Only scan for known vulnerable archives. By default the CLI will scan for class files which are known to be vulnerable which will result in higher signal findings. If you are specifically looking for vulnerable Java archive hashes, use this option.",
 					},
+					&cli.BoolFlag{
+						Name:  "processes",
```
factoidforrest, line 134: I think this description should include the caveat that this only works on linux.  We have a LOT of macos and windows users.

factoidforrest: and maybe the errors that throw in regards to failing to find proc/fs should mention the OS limitation also

factoidforrest (approved):
LGTM

PR #430, opened 2022-02-10
Lunatrace Infrastructure

breadchris:
Lunatrace is deployed using the aws cdk as an ecs task on fargate with a load balancer and consists of:

- backend
- hasura
- oathkeeper
- kratos (wip)


lunatrace/bsl/backend/src/vulnerabilities/handler.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { GrypeVuln, Convert as VulnCast } from '../types/grypeVulnerabilities';
+import { GrypeMeta, Convert as MetaCast } from '../types/grypeVulnerabilityMetadata';
+import { DbType, initDb } from '../utils/vulnerability-loader';
+
+let db: DbType | null = null;
+
+void (async () => {
+  const grypeDbPath = path.join(os.homedir(), '.cache/grype/db/3/vulnerability.db');
+  db = await initDb(grypeDbPath);
```
factoidforrest, line 28: I think id probably synchronously return this promise and then await it inside the functions that need to use it.  Those are async functions anyway, and it ends up a bit cleaner than checking if the db exists and failing somewhat silently with []

factoidforrest: `const db = await dbPromise`

factoidforrest: an anymous top level async function kind of reeks but I understand how we got here.  It would be nice if we could figure out how to turn top-level await on, I know it IS possible

lunatrace/bsl/backend/src/app.ts
```diff
@@ -11,11 +11,30 @@
...
+
 const app = Express();
+app.use(cors());
+app.use(Express.json());
+
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
+app.get('/api/upload-sbom', generatePresignedUrl);
+
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
+app.post('/api/package/vulns', findVulnsForPackages);
```
factoidforrest, line 30: Whos calling this endpoint?  Shouldn't we be using our data that has been imported into hasura/postgres?

factoidforrest (approved):

lunatrace/bsl/backend/src/vulnerabilities/handler.ts
factoidforrest: If this is really just throwaway code then nevermind

lunatrace/bsl/backend-cdk/package.json
```diff
@@ -10,27 +10,37 @@
...
-    "@aws-cdk/aws-route53": "1.138.1",
-    "@aws-cdk/aws-route53-targets": "1.138.1",
-    "@aws-cdk/aws-s3": "1.138.1",
-    "@aws-cdk/core": "1.138.1",
+    "@aws-cdk/assets": "1.142.0",
+    "@aws-cdk/aws-apigateway": "1.142.0",
+    "@aws-cdk/aws-certificatemanager": "1.142.0",
+    "@aws-cdk/aws-ec2": "1.142.0",
+    "@aws-cdk/aws-ecr-assets": "1.142.0",
+    "@aws-cdk/aws-ecs": "1.142.0",
```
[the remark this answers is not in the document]
breadchris: doesn't matter that much, we will not be publishing this package at all. but it isnt crazy to move them over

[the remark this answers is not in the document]
factoidforrest: Ironically it can help with vulnerability scanning LOL. But yeah it doesn't really matter

factoidforrest:
Uh oh linty no pass @breadchris 
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

Uhhhh

PR #432, opened 2022-02-17
Luntrace Login

breadchris:
The login uses ory kratos to store identities and uses oathkeeper as an auth proxy.

An auth context provider is available for getting session information in the dom:
```
const auth = useAuth();
console.log(auth.state.session);
```

email and password as well as github oauth will be supported for the initial release of the login.

Oathkeeper is now meant to be run outside of docker so that it can more easily proxy to services that are in a contain or running on the host. This also makes writing oathkeeper rules more simple since `localhost` is used as the host for both local dev and when running in production to proxy to different services.


lunatrace/bsl/docker-compose.yaml
```diff
@@ -48,8 +48,6 @@ services:
     image: hasura/graphql-engine:v2.1.1
     ports:
     - "8080:8080"
-    extra_hosts:
```
factoidforrest, line 47: Hmm how are we going to do it without this?

factoidforrest, line 47: Hmm how are we going to do it without this?

PR #440, opened 2022-03-08
Adds JWT auth for Token-based auth

freeqaz:
This commit uses the Oathkeeper functionality to "authenticate" inbound requests
by performing an HTTP request to the Backend Node API.

During this request, the Bearer token is forwarded to the specified endpoint and
it's value is used to authenticate the request. If successful, the request has
data added to the JWT for use by Hasura.


factoidforrest (approved):
Noice

PR #444, opened 2022-03-11
Maybe fix the tests

freeqaz:
Bumps the CDK version which is what was breaking the integration tests before.


factoidforrest:
oops i didnt know you were also working on this branch still. well..this contains both of our changes now

PR #446, opened 2022-03-14
Project secret management

factoidforrest:
Secret management controls for projects.


lunatrace/bsl/frontend/src/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface ConfirmationDialogProps {
+  onClose: (success: boolean) => void;
+  show: boolean;
+  title?: string;
+  body?: string;
+}
+
+export const ConfirmationDailog: React.FC<ConfirmationDialogProps> = ({ onClose, show, title, body }) => {
+  return (
+    <>
```
freeqaz, line 26: I am curious -- what does this blank element do when you only have one child? Is it necessary here or is it extraneous?

lunatrace/bsl/frontend/src/hooks/createLogoutHandler.tsx
```diff
@@ -38,7 +38,10 @@ export function createLogoutHandler(navigate: NavigateFunction) {
   return async () => {
     const resolvedLogoutToken = await logoutToken;
     if (resolvedLogoutToken) {
-      void ory.submitSelfServiceLogoutFlow(resolvedLogoutToken).then(() => navigate('/account/login'));
+      void ory.submitSelfServiceLogoutFlow(resolvedLogoutToken).then(() => {
```
freeqaz, line 41: Why `void` this here when you already have an `await` up above for `logoutToken`? If this were to fail for some reason, this currently would be silent. It would be nice to catch this (and maybe do something like make a dialogue box?)

freeqaz (approved):
Nice, this is very solid!

lunatrace/bsl/frontend/src/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Modal show={show} onHide={() => onClose(false)}>
+        <Modal.Header closeButton>{title ? <Modal.Title>{title}</Modal.Title> : null}</Modal.Header>
+        {body ? <Modal.Body>{body}</Modal.Body> : null}
```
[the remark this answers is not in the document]
factoidforrest: thats way better thank you

lunatrace/bsl/frontend/src/hooks/createLogoutHandler.tsx
factoidforrest: word, not my code but i will fix, good catch!

lunatrace/bsl/frontend/src/components/ConfirmationDialog.tsx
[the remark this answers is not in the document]
factoidforrest: i made a HOC to handle this, i just hate how react does it.  Now we have a ConditionallyRender component in utils

lunatrace/bsl/frontend/src/components/ConfirmationDialog.tsx
factoidforrest: yes, definitely.  Theyre harmless though and I kind of like just having them there already when I want to render multiple elements from a component.  idk, im not sure if its harmful.

PR #452, opened 2022-03-16
GitHub App Populates User's Orgs and Projects

breadchris:
The LunaTrace Github App, on install, will populate an organization's repositories in hasura.
When a user login happens, there is a webhook `/api/webhook/github/login` that will get fired and request the user's list of authed orgs and update hasura to reflect this list.


lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
+        },
+      },
+    };
+  }, {} as OrganizationInputLookup);
+
+  await hasura.CreateOrganizations({
+    objects: Object.values(organizations),
+  });
+
+  res.status(302).redirect('http://localhost:4455/');
```
freeqaz, line 117: Nit: Make this dynamic so that it works in production too.

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
 
-export interface KratosIdentityConfig {
-  providers: KratosIdentityProvider[];
-}
+githubApiRouter.get('/webhook/github/install', async (req, res) => {
+  const installationIdQueryParam = req.query.installation_id;
+  const setupActionQueryParam = req.query.setup_action;
+
+  const error = req.query.error;
+  const error_description = req.query.error_description;
```
freeqaz, line 43: Nit: Use camelCase for variables in TS

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
-  initial_refresh_token: string;
-}
+  if (!installationIdQueryParam || typeof installationIdQueryParam !== 'string') {
+    res.status(401).send({
+      error: 'installation_id not provided in query params',
+    });
+    return;
+  }
+
+  const installationId = parseInt(installationIdQueryParam, 10);
```
freeqaz, line 62: Nit: You should wrap this in a `try/catch` block or else these exceptions will crash the server (and possibly hang the request)

freeqaz: What I usually do in order to reduce nesting is to write a helper function like this:
```ts

interface ParsedInt {
  success: true,
  value: number
};

interface ParsedIntError {
  success: false,
  error: string
}

// By using a union type like this, you're able to avoid possibly "undefined" valued based on if
// success is set to true/false. It's a nice TypeScript feature.
function tryParseInt(value, radix = 10): ParsedInt | ParsedIntError {
  try {
    return {
      value: parseInt(value, radix),
      success: true
    };
  } catch (e) {
    return {
      success: false,
      error: e
    };
  }
}

const {success, value} = tryParseInt('123');

if (!success) {
  res.status(401, {error: 'bad value for installation_id', success: false});
}
```

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
-export interface KratosIdentityProvider {
-  initial_id_token: string;
-  subject: string;
-  provider: string;
-  initial_access_token: string;
-  initial_refresh_token: string;
-}
+  if (!installationIdQueryParam || typeof installationIdQueryParam !== 'string') {
+    res.status(401).send({
+      error: 'installation_id not provided in query params',
```
freeqaz, line 57: Nit: I like to return `success: false` in all of my errors because I find it's easier to consume the endpoint.

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
+
+  const installationId = parseInt(installationIdQueryParam, 10);
+
+  const installationData = await pullDataForInstallation(installationId);
+
+  const {
+    data: { repositories },
+  } = installationData;
+
+  type OrganizationInputLookup = Record<string, Organizations_Insert_Input>;
```
freeqaz, line 70: Nit: This could likely benefit from being pulled out into another file. The `reduce` below feels pretty self-contained and potentially useful elsewhere in the app. Also, it clutters this "controller" function a bit too much because it's effectively business logic and not just request/response value mapping.

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -11,42 +11,111 @@
...
+        installation_id: installationId,
+        github_id: organizationId,
+        projects: {
+          data: [...getExistingProjects(orgLookup), project],
+        },
+      },
+    };
+  }, {} as OrganizationInputLookup);
+
+  await hasura.CreateOrganizations({
```
freeqaz, line 113: Is it possible for this to fail? If so, does that need to be handled?

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -68,48 +138,93 @@ function getGithubAccessToken(identity: Identity) {
...
+  let moreDataAvailable = true;
+
+  while (moreDataAvailable) {
+    const userOrgs: GetUserOrganizationsQuery = await github.GetUserOrganizations({
+      orgsAfter: orgsAfter,
+    });
+
+    allUserOrgs = deepmerge(allUserOrgs || {}, userOrgs);
+
+    if (!userOrgs.viewer) break;
```
freeqaz, line 179: Nit: Use braces instead of one liners (Apple once had a bad bug because of this syntax once, so I try to use braces now 100% of the time. We should really just make the linter do this for us.)

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -68,48 +138,93 @@ function getGithubAccessToken(identity: Identity) {
...
-      res.send({
-        error: e.message,
-      });
-    } else {
-      // TODO (cthompson) what other error types exist?
-      res.send({
-        error: e,
-      });
-    }
+  let orgsAfter: string | null | undefined = undefined;
```
freeqaz, line 168: Should this be split out into another function? This seems generally like a useful function `getGitHubOrgsForUser(userId)`

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -68,48 +138,93 @@ function getGithubAccessToken(identity: Identity) {
...
+  }
+
+  const orgNodes = allUserOrgs.viewer.organizations.nodes;
+
+  if (!orgNodes) {
+    throw new Error('organization nodes are not defined');
+  }
+
+  const githubOrgIds = orgNodes.reduce<number[]>((filtered, org) => {
+    if (!org) return filtered;
```
freeqaz, line 200: Nit: use braces

lunatrace/bsl/backend/src/github/projects.ts
```diff
@@ -68,48 +138,93 @@ function getGithubAccessToken(identity: Identity) {
...
+    const orgPageInfo = userOrgs.viewer.organizations.pageInfo;
+
+    orgsAfter = orgPageInfo.hasNextPage ? orgPageInfo.startCursor : null;
+    moreDataAvailable = !!orgsAfter;
   }
+
+  console.log('got orgs for user');
+
+  if (allUserOrgs === null || !allUserOrgs.viewer) {
+    throw new Error('github user organizations response is null');
```
freeqaz, line 190: Nit: If you're going to throw like this, you need to catch it somewhere or else this will dump crap on the consumer of the API.

When you rip this function out to fetch GitHub orgs for a user, I would recommend wrapping your call to that function in a `try/catch` block inside of the Express Controller. If it catches anything then you can return a standard 

```ts
console.error('Failed to fetch GitHub org data for user:', e);
res.status(500).send({error: true, message: 'bad response from GitHub API.');
```

lunatrace/bsl/frontend/src/pages/home.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import React from 'react';
+import { Button, Card, Col, Container, Row } from 'react-bootstrap';
+import { Helmet } from 'react-helmet-async';
+
+export const Home: React.FunctionComponent = (_props) => {
```
freeqaz, line 18: This is tight :ok_hand:

freeqaz (commented):
Here is a bunch of feedback. I'd recommend that we go through and clean some of this up to keep this UX consistent and make it easier for us to keep logs about any possible failures that we see.

Overall, this is awesome and is going to be majorly valuable for making this functionality work!

lunatrace/bsl/backend/src/github/projects.ts
factoidforrest: Good catch I just set this in the linter.  Luckily its always auto fixable

PR #470, opened 2022-03-17
Move LunaDefend into dedicated folder

freeqaz:
This works for me locally. I'm able to run a full Docker build and then pass the integration tests. So I'm going to throw this over the fence and see if the CI agrees with me!


freeqaz:
Closing this because we need to do this piecemeal to figure out what's causing the tests to fail.

PR #474, opened 2022-03-18
node-ipc malware blog post

breadchris:
node-ipc has had some malware pushed to it and this relates to the mission for lunasec to identify and protect against dependencies like this.


factoidforrest:
nice one! ill do a little editing pass if its alright with you, just to give more context sooner in the article for readers, before drilling down.  We kind of do the opposite here, by drilling down and then coming back up, but that might be harder for a lot of people hitting this topic who haven't seen it before.

PR #484, opened 2022-03-25
Vuln ignore

factoidforrest:
Now possible to ignore findings in bulk and individually.  Not possible to un-ignore


factoidforrest:
why no actions?

freeqaz (approved):
Looks great! This will be huge to have.

PR #493, opened 2022-03-28
SBOM generation for repositories

breadchris:
SBOMs are now generated for repositories and get forwarded along the existing ETL

When an SBOM is scanned for vulnerabilities, the build is checked for the presence of a pull request id. If there is one present, then a comment on the pull request will be made.


freeqaz (approved):
Definitely a little too much code to keep it all in my head, but everything here is conceptually in line with my understanding. I didn't realize the CLI changes would be in this branch too, but it all makes sense how it's working now. Let's get this landed and deployed!

PR #495, opened 2022-03-29
Add JWKS to JWT auth via Express middleware

freeqaz:
This also changes the default GraphQL Schema generation to happen from Hasura directly instead of adding an intermediary `schema.graphql` for both the front-end and back-end. TypeScript files are still generated and committed to Git though.


lunatrace/bsl/hasura/metadata/databases/lunatrace/tables/public_projects.yaml
```diff
@@ -27,13 +27,6 @@ array_relationships:
       table:
         name: ignored_vulnerabilities
         schema: public
-- name: github_repositories
```
freeqaz, line 30: These were duped and the metadata was unable to be applied. I'm not sure exactly why -- probably a bad merge -- but this fixes it to remove the dupes. (Expand the file to verify. The "dupes" are 100% the same.)

factoidforrest (approved):

PR #500, opened 2022-03-31
Add hotfix patch to the doc

blindpirate:


factoidforrest:
Wrote it in a different way as part of a bigger edit but thanks for the PR!

PR #518, opened 2022-04-06
Fix up the report generator logic and add common module

freeqaz:
There are some Hasura changes that are a bit tricky to navigate, just FYI


freeqaz:
Oof I'll have to hide the table when there are zero findings. I thought I did but apparently my logic is broken!

factoidforrest (commented):
TEST COMMENT

PR #519, opened 2022-04-06
Better git info in projects and snapshots... and fix SQS jobs

factoidforrest:
WIP, appears to be making the lunatrace CLI segfault after I merged the changes between chris and I.  Worked before that, will fix before merge


lunatrace/bsl/frontend/codegen.yml
```diff
@@ -6,7 +6,21 @@ documents: "src/api/graphql/**/*.graphql"
 generates:
   ./src/api/generated.ts:
     plugins:
-      - "typescript"
```
freeqaz, line 9: This is very hype

freeqaz (approved):
It looks like you were flying today! I'll have to merge these with my changes tomorrow. This is very slick 🙏

factoidforrest:
Okay @freeqaz I did the zero downtime pattern, I think it will be fine. Just do the first two of these three new migrations but NOT the last one until the new cluster is up, then the last one is a cleanup step.

lunatrace/bsl/hasura/migrations/lunatrace/1649108103633_rename-api-response-to-traits-and-add-build-type/up.sql
```diff
@@ -0,0 +1,25 @@
+
+ALTER TABLE public.github_repositories ADD COLUMN "traits" jsonb NULL;
+UPDATE public.github_repositories SET traits = api_response WHERE traits IS NULL;
```
factoidforrest, line 3: heres the magic

PR #524, opened 2022-04-09
filtering comment findings now, still not editing existing comment

factoidforrest:
We are now using Reviews in "comment" mode instead of straight comments.  They are updated instead of duplicated.  

shot_031

@freeqaz I rolled back most of the changes you made when you copy pasted the code, and then for the backend used hasura's return data from the `InsertScan` mutation instead of the input data because the shape was much more similar to what the frontend was seeing, and we could re
quest IgnoredVulnerabilities simultaneously which we needed.


lunatrace/bsl/common/build/main/group-vulnerabilities-by-package.js
```diff
@@ -0,0 +1,80 @@
...
+ * License. You may obtain a copy of the License at
+ *
+ * https://github.com/lunasec-io/lunasec/blob/master/licenses/BSL-LunaTrace.txt
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+const types_1 = require("./types");
+function sortBySeverity(a, b) {
```
freeqaz, line 18: nuke this file please (the whole `build` folder)

lunatrace/bsl/common/src/logger/stack-trace.ts
```diff
@@ -0,0 +1,36 @@
...
+// because there was something wrong with that module
+
+export function get(belowFn: (...args: any[]) => unknown) {
+  const oldLimit = Error.stackTraceLimit;
+  Error.stackTraceLimit = Infinity;
+
+  const dummyObject = {};
+
+  const v8Handler = Error.prepareStackTrace;
+  Error.prepareStackTrace = function (dummyObject, v8StackTrace) {
```
freeqaz, line 24: Nit that we should add to the linter: If you don't use an anonymous function, it should be named.

lunatrace/bsl/frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@aws-sdk/client-s3": "^3.52.0",
-    "@lunatrace/lunatrace-common": "link:../common",
+    "@lunatrace/lunatrace-common": "workspace:~",
```
freeqaz, line 7: the linter seems to be complaining about this. Are you sure this is the correct syntax?

PR #527, opened 2022-04-10
Lunalogger

factoidforrest:
## LunaLogger

This is a slim JSON logger, with an API similar to native console.  It supports pretty colors, pretty json output, and
typescript callsites(probably).

```typescript
import {LunaLogger} from '@lunatrace/lunatrace-common'
const log = new LunaLogger({ trace: true }, {whatever:'extra fields'});
log.addTransport(new JsonTransport({ minLevel: 'debug', colors: false, pretty: true }));

//basic usage
log.info('some log')
```
outputs this but with pretty colors:
```json

{
    "level": "debug",
    "timeEpoch": 1649550906635,
    "message": "some log",
    "timePretty": "Sat Apr 09 2022 17:35:06 GMT-0700 (Pacific Daylight Time)",
    "whatever":"extra fields",
    "loggerName": "default"
}
```
```typescript
//if the first argument is an object, any fields end up on the root logged object instead of in the message, just like pino
log.log({ test: 'field' }); // alias for info
```
```json
    {
      "level": "info",
      "timeEpoch": 1649551029078,
      "message": "",
      "timePretty": "Sat Apr 09 2022 17:37:09 GMT-0700 (Pacific Daylight Time)",
      "loggerName": "default",
      "test": "field"
    }
```

If we want another transport that doesnt print JSON for dev so its more readable, we can add that, or maybe add that option to the existing transport.


factoidforrest:
I'd say this is ready for merge, it seems to be working well.  Maybe we can combine it with @breadchris PR and see if it does the job.  I'd still like to make a non-json based transport for dev since the json is a little clunky but..i think this will work for now.

breadchris (approved):

PR #548, opened 2022-04-18
add migration for cascading an organization removal

breadchris:
Removing an organization does not currently work because foreign keys are blocking its removal. 

Cascading was the chosen method for removing the related objects as an organization being deleted from LunaTrace should result in all of their associated data being removed.


factoidforrest:
test

PR #561, opened 2022-04-22
Move the hasura actions into graphql resolvers

factoidforrest:
Linter still broken!  Need to fix that before merge


freeqaz (approved):
Just fix the linter but this otherwise looks good!

Issue #552, opened 2022-04-20
Do master builds immediately when project is installed and when commits happen


factoidforrest:
connected to #455 since that data will go into the project home

PR #567, opened 2022-04-27
Scan master on install and repo added

breadchris:
When a new installation is performed or a new repo is added, start a scan over all accessible repos on the default branch.

This will hydrate the state of the project page as there will be a scan that has been performed for each.


lunatrace/bsl/backend/src/config.ts
```diff
@@ -77,12 +94,45 @@ export function getEtlBucketConfig(): SbomHandlerConfig {
...
+      queueNameEnvVar: 'PROCESS_MANIFEST_QUEUE'
+    },
+    'process-sbom': {
+      maxMessages: DEFAULT_QUEUE_MAX_MESSAGES,
+      visibility: DEFAULT_QUEUE_VISIBILITY,
+      queueNameEnvVar: 'PROCESS_SBOM_QUEUE'
+    },
+    'process-repository': {
+      maxMessages: DEFAULT_QUEUE_MAX_MESSAGES,
+      visibility: DEFAULT_QUEUE_VISIBILITY,
```
freeqaz, line 121: You need to make this a longer time period than 60 seconds. You can see above on line 106 that I have this set to 5 minutes.

How long should it be set to? Well, how long do we expect a Git Clone operation to take?

lunatrace/bsl/backend/src/config.ts
```diff
@@ -77,12 +94,45 @@ export function getEtlBucketConfig(): SbomHandlerConfig {
...
+export function getQueueHandlerConfig(): QueueHandlerWorkerConfig {
   const handlerName = checkEnvVar('QUEUE_HANDLER');
+
+  const DEFAULT_QUEUE_MAX_MESSAGES = 10;
+  const DEFAULT_QUEUE_VISIBILITY = 60;
+
+  const handlerConfigLookup: Record<QueueHandlerType, QueueHandlerConfig> = {
+    'process-webhook': {
+      maxMessages: 1,
+      visibility: DEFAULT_QUEUE_VISIBILITY * 5,
```
freeqaz, line 106: We should probably make this shorter now unless you think this will take a while still.

lunatrace/bsl/backend/src/github/actions/generate-snapshots-for-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+  const installationToken = await getInstallationAccessToken(installationId);
+
+  if (installationToken.error) {
+    log.error('unable to get installation token', {
+      error: installationToken.msg
+    });
+    return;
+  }
+
+  const github = generateGithubGraphqlClient(installationToken.res);
```
freeqaz, line 32: We should just pass this Client in as an argument to this function. The fact that we have to call `getInstallationAccessToken` and then setup the client manually with the `installationToken` is brutal. We're already doing that in a bunch of places in our codebase.

lunatrace/bsl/backend/src/github/actions/generate-snapshots-for-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  await Promise.all(githubRepos.map(async (repo) => {
+    const githubRepo = await catchError(async () => await github.GetRepository({
```
freeqaz, line 35: I'm pretty sure you have a redundant anonymous function here. This can be simplified as:
```ts
const githubRepo = await catchError(github.GetRepository({
  name: repo.repoName,
  owner: repo.orgName
}));
```

The reason is that `github.GetRepository` returns a promise and so the `async () => await` bit is entirely redundant.

lunatrace/bsl/backend/src/github/actions/generate-snapshots-for-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+        repo,
+        githubRepo
+      });
+      return;
+    }
+
+    const cloneUrl = `${githubRepo.repository.url}.git`;
+    const gitBranch = githubRepo.repository.defaultBranchRef.name;
+
+    await generateSnapshotForRepository({
```
freeqaz, line 59: Just to confirm, does this do a shallow Git clone? I see that it calls into the CLI, but I just want to verify.

It might not always be possible. I thiiiiink you probably can't always have a shallow clone because, iirc, you can't "check out a given commit hash" when you've done a shallow clone. If that's true, that means that there is a race condition (you try to checkout the commit that was the "head" but then another push happens while this job is running).

Maybe the workaround then is to just use `--single-branch`? I'm punting this to you to verify though. :pray:

lunatrace/bsl/backend/src/github/actions/generate-snapshots-for-repositories.ts
```diff
@@ -0,0 +1,67 @@
...
+      cloneUrl,
+      gitBranch,
+      repoGithubId: repo.repoId,
+      installationId,
+      sourceType: 'gui',
```
freeqaz, line 64: Forrest here: This sourceType will show a little drag and drop icon in the gui.  Its probably more correct to stick with the github type, since thats where this came from

lunatrace/bsl/backend/src/github/webhooks.ts
```diff
@@ -13,27 +13,34 @@
...
 import {orgMemberAdded} from "./actions/org-member-added";
-import {reviewPullRequest} from "./actions/review-pull-request";
+import {generateGithubGraphqlClient} from "./api";
 import { getInstallationAccessToken } from './auth';
-import {WebhookInterceptor} from './webhook-cache';
-
-const webhookQueue = process.env.PROCESS_WEBHOOK_QUEUE;
+import {WebhookInterceptor} from "./webhook-cache";
+
+export function registerWebhooksToInterceptor(interceptor: WebhookInterceptor): void {
```
freeqaz, line 28: Nice change. That's dope

lunatrace/bsl/backend/src/github/webhooks.ts
```diff
@@ -106,30 +115,17 @@ async function pullRequestHandler(event: EmitterWebhookEvent<'pull_request'>) {
...
     const gitBranch = event.payload.pull_request.head.ref;
     const installationId = event.payload.installation.id;
 
-    await reviewPullRequest({
+    await generateSnapshotForRepository({
+      cloneUrl,
+      gitBranch,
+      repoGithubId,
       installationId,
+      sourceType: 'pr',
```
freeqaz, line 127: This is the right source type for above (not `gui`).

freeqaz (approved):
Looks very dope! Just a handful of things to tweak but this should be good to go. Both @factoidforrest and I reviewed this together. :)

Oh, and you might want to pull in master + run the linter because it'll fix a bunch of your lint errors. We'll be landing that as soon as the CI is happy. :pray:

lunatrace/bsl/backend/src/config.ts
breadchris: this is a good question. I am going to set this to 10 minutes for the moment. It might be too long, but I would rather have the window be on the longer side for the moment until we have metrics around it.

breadchris:
closes #552

PR #577, opened 2022-05-03
Install fixes: Add node-gyp and lock engines to node v16

ajvpot:
Adds `node-gyp` as a dependency of the root workspace so that modules that modules without prebuilt binaries can be built and avoid
```
Usage Error: Couldn't find a script name "node-gyp" in the top-level (used by deasync@npm:0.1.24). This typically happens because some package depends on "node-gyp" to build itself, but didn't list it in their dependencies. To fix that, please run "yarn add node-gyp" into your top-level workspace. You also can open an issue on the repository of the specified package to suggest them to use an optional peer dependency.
```

Pins `engines` to `^16` because `deasync` won't build against node v17. Ask me how I found out.

### DIFF
--- a/.pnp.loader.mjs
+++ b/.pnp.loader.mjs
@@ -1,19 +1,3 @@
-/*
- * Copyright 2022 by LunaSec (owned by Refinery Labs, Inc)
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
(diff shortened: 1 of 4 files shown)


.pnp.loader.mjs
```diff
@@ -1,19 +1,3 @@
-/*
```
ajvpot, line 1: Copyright statement was removed because the file is generated

factoidforrest (approved):
Very smart to lock people off of node 17 if at all possible.  Nice!

factoidforrest:
@ajvpot huh its complaining about commit signing but all your commits are signed?  very odd

PR #579, opened 2022-05-03
Changes for bringing up frontend docs

ajvpot:
* Updates guide for bringing up frontend.
* Removes duplicated instructions in README.md, replaced with a pointer to README_DEV_SETUP.md
* Adds POSTGRES_DB env var to docker compose to create `lunasec` db on startup, this is required before bootstrapping ory/hasura.
* Replaced docker ps grep oneliner with docker compose equivelant.
* Added env vars that were missing

Borrowed some sql migrations from mac bring up diff.

### DIFF
--- a/lunatrace/bsl/hasura/migrations/lunatrace/1649717845749_create-users-table-use-as-auth-table/up.sql
+++ b/lunatrace/bsl/hasura/migrations/lunatrace/1649717845749_create-users-table-use-as-auth-table/up.sql
@@ -21,9 +21,9 @@ ALTER TABLE public.organization_user ADD CONSTRAINT organization_user_user_id_fk
   REFERENCES public.users
   (id) ON UPDATE cascade ON DELETE cascade;
 
-UPDATE users SET github_id = (
+UPDATE public.users SET github_id = (
   SELECT kratos.github_id
   FROM (
-    SELECT config->'providers'->0->>'subject' as github_id, identity_id FROM identity_credentials JOIN identities ON identities.id = identity_credentials.identity_id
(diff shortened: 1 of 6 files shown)


lunatrace/bsl/hasura/migrations/lunatrace/1649717845749_create-users-table-use-as-auth-table/up.sql
```diff
@@ -21,9 +21,9 @@ ALTER TABLE public.organization_user ADD CONSTRAINT organization_user_user_id_fk
...
   (id) ON UPDATE cascade ON DELETE cascade;
 
-UPDATE users SET github_id = (
+UPDATE public.users SET github_id = (
   SELECT kratos.github_id
   FROM (
-    SELECT config->'providers'->0->>'subject' as github_id, identity_id FROM identity_credentials JOIN identities ON identities.id = identity_credentials.identity_id
-  ) as kratos WHERE users.kratos_id = kratos.identity_id AND kratos.github_id IS NOT NULL
-)
+    SELECT config->'providers'->0->>'subject' as github_id, identity_id FROM public.identity_credentials JOIN public.identities ON public.identities.id = public.identity_credentials.identity_id
```
factoidforrest, line 27: Nice fixes, i just want to be careful this won't break anything in the prod db.  Because we have prod data now, we are having to make new migrations for all db changes.

ajvpot: I think this is alright after discussing with Free and Chris, this shouldn't affect the existing schema. Prod has already run this migration so it will not execute it again and this does not change the result. We have a backup at the ready in case we need to restore.

PR #588, opened 2022-05-05
Impersonate a user as an admin

breadchris:
If you are an admin, then you are able to impersonate a user to perform actions on their behalf in the UI.


ajvpot (approved):

lunatrace/bsl/logger/src/index.ts
```diff
@@ -74,26 +116,22 @@ export class LunaLogger {
...
-    const fieldsFromThread = threadStorage ? threadStorage.additionalFields : {};
+    const contextFromStorage = threadStorage ? threadStorage.context : {};
 
     const logObject: LogObj = {
       level,
       timeEpoch: now.getTime(),
+      name: this.options.loggerName || defaultLoggerName,
       message: '',
+      context: {
+        ...contextFromStorage,
```
factoidforrest, line 131: Yeah nice, I approve this message

factoidforrest: Stoked we are taking co-ownership of this little logger :)

lunatrace/bsl/logger/src/tests/logger.test.ts
```diff
@@ -59,7 +59,7 @@ describe('LunaLogger', () => {
 
   describe('child logger', () => {
     it('appends any child additional fields', () => {
-      const childLogger = log.child({ child: 'field' });
+      const childLogger = log.child('child', { child: 'field' });
```
factoidforrest, line 62: Awesome and you ran these? If so, great.

lunatrace/bsl/frontend/src/api/graphql/getSidebarInfo.graphql
```diff
@@ -27,4 +27,10 @@ query GetSidebarInfo {
             }
         }
     }
+
+    users(where: $users_filter) {
```
factoidforrest, line 31: As discussed we are moving this to its own API call and only calling it from within the SideBar.tsx component.

lunatrace/bsl/frontend/src/store/slices/authentication.ts
```diff
@@ -85,6 +87,20 @@ export const selectRegisterFlow = (state: RootState) => state.auth.registerFlow;
 export const selectUserId = (state: RootState) => state.auth.session?.identity.id;
 export const selectConfirmedUnauthenticated = (state: RootState) => state.auth.confirmedUnauthenticated;
 
+const sidebarInfoResult = api.endpoints.GetSidebarInfo.select();
```
factoidforrest, line 90: As discussed we are completely getting rid of this function and moving it into either: The sidebar.tsx component itself or a helper

lunatrace/bsl/frontend/src/utils/users.ts
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { GetSidebarInfoQuery } from '../api/generated';
+import { impersonateUserKey } from '../constants/localstorage';
+import { ImpersonateUser } from '../types/user';
+
+export function getImpersonatedUser(): ImpersonateUser | null {
+  const user = localStorage.getItem(impersonateUserKey);
```
factoidforrest, line 19: This is ok but we actually already have the code to do this in the app, and its a bit better because it uses a hook to do it:
https://github.com/lunasec-io/lunasec/blob/6fc6fa0e08234f0032ab570b496eec4c2830e53d/lunatrace/bsl/frontend/src/hooks/useLocalStorage.ts#L16

You can see this strategy being used from some of our contexts/providers as well.  If you only use it in one place, then you can just call that hook directly from the component that needs it.  If you want to use it in multiple places and have it rereder automatically on change, then the way to do that is with a provider. In that case, you would write a context called `ImpersonatedUser` that provides this data and is responsible for mapping it into localStorage using the `useLocalStorage` hook.  Then, any changes will cascade through the app/DOM automatically, without a component having to know to reload this information out of localStorage.  Thats only if you want to share this state app-wide. You can see an example with this context provider https://github.com/lunasec-io/lunasec/blob/master/lunatrace/bsl/frontend/src/contexts/LayoutContext.tsx

So yeah, TLDR, try to use the machinery we have and do this the 'reacty' way.

lunatrace/bsl/logger/src/tests/logger.test.ts
factoidforrest: I just ran these and theyre failing.

lunatrace/bsl/hasura/metadata/databases/lunatrace/tables/public_builds.yaml
```diff
@@ -1,174 +1,174 @@
 table:
```
factoidforrest, line 1: When applying metadata I am getting: 
```╰─$ hasura metadata apply 
FATA[0001] error applying metadata 
{
  "path": "$.args.metadata.remote_schemas[0].permissions[0].definition.schema",
  "error": "parsing the schema document: \"type definition: string\"",
  "code": "parse-failed"
} 
```
Does it work for you if you do hasura metadata apply from this branch?

lunatrace/bsl/frontend/src/components/navbar/Navbar.tsx
```diff
@@ -56,6 +58,11 @@ const NavbarComponent: React.FunctionComponent<NavbarComponentProps> = ({ setupW
     </span>
   );
 
+  const stopImpersonating = () => {
+    setImpersonatedUser(null);
+    window.location.reload();
```
factoidforrest, line 63: This really isnt nuts but doing the context/provider thing I mentioned might eliminate this altogether.  Not saying a reload is a bad idea.

lunatrace/bsl/frontend/src/components/sidebar/sidebarItems.tsx
```diff
@@ -11,17 +11,21 @@
...
-import { GetSidebarInfoQuery } from '../../api/generated';
+import { GetSidebarInfoQuery, Scalars } from '../../api/generated';
+import useAppSelector from '../../hooks/useAppSelector';
+import { userIsAdmin } from '../../store/slices/authentication';
 import { userHasAnyOrganizations } from '../../utils/organizations';
 
 import { NavSection, SidebarItem } from './types';
 
 export function generateSidebarItems(data: GetSidebarInfoQuery | undefined, isAuthenticated: boolean): NavSection[] {
+  const isAdmin = useAppSelector(userIsAdmin);
```
factoidforrest, line 27: We talked about getting this out of the store and instead just passing a boolean to this function `isAdmin`

lunatrace/bsl/frontend/src/pages/account/admin.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { impersonateUserHeader } from '../../constants/headers';
+import { ImpersonateUser } from '../../types/user';
+import { inputChangeHandler } from '../../utils/input-helpers';
+import { setImpersonatedUser } from '../../utils/users';
+
+export const AdminDashboard = () => {
+  const navigate = useNavigate();
+  const [username, setUsername] = useState<string>('');
+
+  const { data, error, isLoading } = api.endpoints.GetLunaTraceUsers.useQuery();
```
factoidforrest, line 28: That's interesting.  They also should be populating hooks for us onto the api object so you can do 
`api.useGetLunaTraceUsersQuery` and it should work. Idk, maybe you doing it this way has to do with the type struggles you were having.

lunatrace/bsl/frontend/src/pages/project/dashboard/Main.tsx
```diff
@@ -48,9 +65,28 @@ export const ProjectDashboardMain: React.FunctionComponent<ProjectDashboardMainP
...
-  console.log('project info is ', project);
+  const cloneProject = () => {
+    const cloneUrl = project.github_repository?.authenticated_clone_url?.url;
+    if (!cloneUrl) {
+      return null;
+    }
+    const formattedCloneUrl = cloneUrl.replace('git://', 'https://');
+    return (
+      <>
+        <CopyBlock
```
factoidforrest, line 76: very, very nice

factoidforrest: Thanks for reusing that lib I already brought in too.

factoidforrest (approved):
Cool! I definitely requested some hefty changes around react state management, but otherwise super solid.

PR #590, opened 2022-05-05
Fix broken things, particularly in the frontend

factoidforrest:
Issues resolved by this PR: 

- Frontend not working on mobile because sidebar locked open, blocking viewport
- Personal project generation no longer firing due to change to webhook user registration, making the wizard link broken for new users.  Personal project is now opt-in by clicking a link in the wizard
- Sidebar being completely inaccessible for any guest users, preventing access to the public databases

Other optimizations:

- Wizard status now lives properly in a react context, instead of prop drilling from layout
- x-hasura-real-user-id is now accessible in hasura and corresponds to the real user id of the logged in user, not the kratos id
- identity -> user is now a 1:1 relation in hasura and is UNIQUE()
- tmux launching docker compose is now unbroken for users using text passwords


CODEOWNERS
```diff
@@ -0,0 +1,70 @@
+
+
+# Forrest
+lunatrace/bsl/frontend @factoidforrest
+lunatrace/bsl/backend/graphql-yoga @factoidforrest
+lunatrace/bsl/common @forrest
```
breadchris, line 6: @forrest needs to be @factoidforrest

breadchris (changes requested):

CODEOWNERS
factoidforrest: thanks meng

breadchris (changes requested):

breadchris (approved):

lunatrace/bsl/frontend/src/contexts/SidebarContext.tsx
```diff
@@ -34,7 +34,7 @@ const initialState = {
 const SidebarContext = React.createContext(initialState);
 
 function SidebarProvider({ children }: { children: React.ReactNode }) {
-  const [isOpen, setIsOpen] = useState(false);
+  const [isOpen, setIsOpen] = useState(window.innerWidth > 992);
```
ajvpot, line 37: super nit: can we use a media query for this or something?

ajvpot (approved):

PR #591, opened 2022-05-05
Download SBOM button

### LINKED ISSUE #467: Download Sbom button on build

ajvpot:
Adds a button for downloading the SBOM for any given snapshot.

Closes #467.


ajvpot:
image

lunatrace/bsl/hasura/metadata/databases/lunatrace/tables/public_builds.yaml
```diff
@@ -27,6 +27,16 @@ array_relationships:
...
         name: scans
         schema: public
+remote_relationships:
+- definition:
+    hasura_fields:
+    - id
+    remote_field:
+      SBOMURL:
+        arguments:
+          buildId: $id
```
factoidforrest, line 37: very nice use of this hasura feature.  This definitely works as is but I probably would have taken a JIT approach where clicking the "download sbom" button actually goes and presigned the url and then fires it.  

As you have it, the data is eager loaded onto the object whenever this "BuildDetails" call happens.  If this was just an internal route, then it would just be a little wasteful and NBD, but since this is calling the external AWS API every time, I personally would separate this out into its own call that the frontend makes when it needs to.  After all, people are probably going to click the "download" button 1/100 times they load the build details page.

Another option, if you're comitting to leaving it as you have it, would be to not request that field until you actually need it and then fire a new query that DOES include the field.  You can do this with useLazyQuery in RTK.  Id probably just make it a separate direct call to the remote schema as described above though.  Either way, useLazyQuery combined with `unwrap` will do the trick. You can see me doing something similar for the manifest upload drag-and-drop flow.

factoidforrest (approved):
Nice!  I definitely requested some big changes, your discretion on if we do them now or later.

lunatrace/bsl/backend/src/graphql-yoga/resolvers/sbom-url.ts
```diff
@@ -0,0 +1,42 @@
...
+import { checkProjectIsAuthorized } from './presign-manifest-upload';
+
+type SBOMURLResolverT = NonNullable<QueryResolvers['SBOMURL']>;
+
+export const SBOMURLResolver: SBOMURLResolverT = async (parent, args, ctx, info) => {
+  throwIfUnauthenticated(ctx);
+  const build = await hasura.GetBuild({ build_id: args.buildId });
+  await checkProjectIsAuthorized(build.builds_by_pk?.project?.id, ctx);
+
+  let downloadURL = '';
```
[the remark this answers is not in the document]
ajvpot: I can change it but is there a reason for that?

lunatrace/bsl/backend/src/graphql-yoga/schema.graphql
```diff
@@ -9,6 +9,9 @@ type Query {
         orgId: uuid!
         buildId: uuid!
     ): SbomUploadUrlOutput
+    SBOMURL(
```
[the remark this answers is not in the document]
ajvpot: I was doing SBOMURL because they're both acronyms. I agree about consistency, will change it to sbomURL.

lunatrace/bsl/hasura/metadata/databases/lunatrace/tables/public_builds.yaml
ajvpot: I don't think we're calling AWS here. Signing a URL uses the key and can be done offline. I could still make it lazy but I think this is acceptable because it doesn't make any external calls. WDYT?

freeqaz: I think it's probably worth moving this to be "lazy" simply because it will prevent a "stale link" from getting clicked if somebody has the page open for >10 minutes.

There is an argument to be made around rate limiting, as well (if we ever wanted to limit how many SBOMs could be downloaded from S3), but that's less important. I think the UX concern is far more important.

PR #614, opened 2022-05-12
query less data when loading project details

### LINKED ISSUE #609: Deal with build/finding overfetching

factoidforrest:
fixes #609


lunatrace/bsl/common/src/group-vulnerabilities-by-package.ts
```diff
@@ -98,3 +98,18 @@ function mergeExistingFindingIntoPackage<F extends Finding>(vulnPackage: Vulnera
     });
   }
 }
+
+interface FindingForCounting {
```
ajvpot, line 102: Would it be practical to use hasura aggregation functions to return the count grouped by severity instead of returning all keys?

factoidforrest: yes but its extremely tricky so not right now.  We need to filter by ignored, which is what makes the query more difficult. I tried to get that working in a postgres function and gave up because it was really tricky.  

We should definitely come back and do that.

ajvpot (approved):

PR #617, opened 2022-05-12
combine sqs queue handlers

breadchris:
Having so many queue handlers has been helpful for speedy development of adding new queue sources, but it has made development and traceability of issues difficult. Combing the queue handlers will help us during local development and simplify our production infrastructure.


lunatrace/bsl/backend/src/config.ts
```diff
@@ -124,22 +124,33 @@ export function getQueueHandlerConfig(): QueueHandlerWorkerConfig {
...
+    // TODO (cthompson) fill these in
+    's3-queue-handler': {
       maxMessages: DEFAULT_QUEUE_MAX_MESSAGES,
-      visibility: DEFAULT_QUEUE_VISIBILITY,
-      envVar: 'PROCESS_SBOM_QUEUE',
+      visibility: DEFAULT_QUEUE_VISIBILITY * 10,
+      envVar: 'PROCESS_REPOSITORY_QUEUE',
     },
-    'process-repository': {
+    'lunatrace-queue-handler': {
```
factoidforrest, line 153: And i guess the implication here is that lunatrace-queue-handler takes events that we have fired, instead of fired from s3?  I would probably call it custom-event-queue-handler or something. Just a little confused because the whole thing is  lunatrace isn't it?  

Not saying we have to change it

lunatrace/bsl/backend/src/types/sqs.ts
```diff
@@ -25,7 +25,25 @@ export interface QueueHandlerConfig {
 
 export type BuildSourceType = 'pr' | 'gui' | 'cli';
 
-export interface GenerateSnapshotForRepositoryRecord {
+export type LunaTraceSqsEventType = 'repository-snapshot' | 'process-webhook';
```
factoidforrest, line 28: `type LunaTraceSqsEventType = LunaTraceSqsEvent['type']` If you want to be fancy.

factoidforrest (approved):
nice

lunatrace/README_DEV_SETUP.md
```diff
@@ -83,12 +80,9 @@ arn:aws:cloudformation:us-west-2:134071937287:stack/lunatrace-alex-EtlStorage/46
 You'll need to format those values into an env file at `$REPO/lunatrace/dev-cli/.env.dev` that looks like:
 
 ```env
-S3_SBOM_BUCKET=lunatrace-alex-etlstorage-sbombucket8550fee8-1drqtwb7yf2dg
-S3_MANIFEST_BUCKET=lunatrace-alex-etlstorage-manifestbucket46c412a5-1b0kc5cys81er
-PROCESS_WEBHOOK_QUEUE=lunatrace-alex-EtlStorage-ProcessWebhookProcessingQueue475F8047-ZzRXj42PsJAZ
-PROCESS_REPOSITORY_QUEUE=lunatrace-alex-EtlStorage-ProcessRepositoryProcessingQueueD69CAAE0-ma9VglOAlgiF
-PROCESS_MANIFEST_QUEUE=lunatrace-alex-EtlStorage-ProcessManifestProcessingQueue7A1F1CB2-BWVszydikZnB
-PROCESS_SBOM_QUEUE=lunatrace-alex-EtlStorage-ProcessSbomProcessingQueueA3A9FE69-6CG55qoQkM7q
+S3_SBOM_BUCKET=xxx
```
factoidforrest, line 83: way smarter

lunatrace/bsl/backend/src/workers/queue-processors/process-luna-trace-sqs-message.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+import { createGithubWebhookInterceptor } from '../../github/webhooks';
+import { LunaTraceSqsMessage, MessageTypeToActivityLookup } from '../../types/sqs';
+import { MaybeErrorVoid } from '../../types/util';
+import { log } from '../../utils/log';
+import { processGithubWebhookActivity } from '../activities/process-github-webhook-activity';
+import { snapshotRepositoryActivity } from '../activities/snapshot-repository-activity';
+
+export async function createMessageTypeToActivityLookup(): Promise<MessageTypeToActivityLookup> {
```
factoidforrest, line 21: ok so activities = handlers

factoidforrest: If I wrote this, this would be `createQueueHandlers`.  This isn't a dynamically generated lookup table, its a flat object that has the instantiated queue handlers in it

lunatrace/bsl/backend/src/workers/queue.ts
```diff
@@ -0,0 +1,190 @@
...
+      const data = await sqsClient.send(new DeleteMessageCommand(deleteParams));
+      log.info('Message deleted', data);
+    } catch (err) {
+      log.error('Error deleting message', err);
+    }
+  }
+
+  // processMessageAndReturnResults determines how to process a message from a queue
+  async processMessageAndReturnResults(parsedMessage: any): Promise<MaybeErrorVoid[]> {
+    if ('type' in parsedMessage) {
```
factoidforrest, line 91: Are we trusting amazon to not also use that key and break us?  Maybe its fine but a glance it looks brittle. Should we maybe include a special key like `_isLunasecMessage: true`? Nice comments

lunatrace/bsl/backend-cdk/bin/lunatrace-backend.ts
```diff
@@ -87,9 +87,12 @@ function deployStack() {
     }
 
     const publicBaseUrl = `http://localhost:4455`;
+    const development = true;
+
     return new WorkerStorageStack(app, `${appName}-${devUser}-EtlStorage`, {
       env,
       publicBaseUrl,
+      development,
```
factoidforrest, line 95: define these straight in the object

lunatrace/bsl/backend/src/workers/queue.ts
```diff
@@ -0,0 +1,190 @@
...
+      // if the key 'type' is present in the message, then this message is one that we have sent from one of our services
+      if (this.activityLookup === null) {
+        throw new Error('activityLookup is null. Make sure that setup was called first.');
+      }
+      return await processLunaTraceSqsMessage(this.activityLookup, parsedMessage);
+    } else if ('Records' in parsedMessage) {
```
factoidforrest, line 97: I guess thats all they really give us. The records have "eventSource":"aws:s3",
But its in the array

lunatrace/bsl/backend/src/workers/queue.ts
```diff
@@ -0,0 +1,190 @@
...
+    try {
+      const data = await sqsClient.send(new DeleteMessageCommand(deleteParams));
+      log.info('Message deleted', data);
+    } catch (err) {
+      log.error('Error deleting message', err);
+    }
+  }
+
+  // processMessageAndReturnResults determines how to process a message from a queue
+  async processMessageAndReturnResults(parsedMessage: any): Promise<MaybeErrorVoid[]> {
```
factoidforrest, line 90: `callHandlerBasedOnMessageSource`

lunatrace/dev-cli/src/cli.ts
```diff
@@ -25,53 +25,29 @@ dotenv.config();
...
+  queueWorker,
   smeeWebhook,
-  webhookWorker,
 } from './services';
 import { tmuxpConfig, tmuxWindow } from './tmux';
 
 const servicesWindow = tmuxWindow('services', [hasura, frontend, dockerCompose]);
 
-const workerWindow = tmuxWindow('workers', [manifestWorker, sbomWorker, webhookWorker, repositoryWorker]);
+const workerWindow = tmuxWindow('workers', [queueWorker]);
```
factoidforrest, line 43: YEYYY

factoidforrest (approved):

lunatrace/bsl/backend/src/workers/queue-processors/process-luna-trace-sqs-message.ts
breadchris: I am specifically trying to avoid "handler" anywhere in the queue code because that term is very overloaded. I chose "activity" because this is the term that is used in Temporal to define a "single, well defined action" https://docs.temporal.io/concepts/what-is-an-activity/. If/when we chose to use temporal, all of the queue code will go away, but the activities, as they are defined now, will not need to be changed at all.

lunatrace/bsl/backend/src/workers/queue.ts
breadchris: Interesting point, but if `type` is being used by another amazon data source, then that still works in this design. The person hooking up with data source would add another type to the list of types that are recognized by the message processor and everything should work as expected.

lunatrace/bsl/backend/src/workers/queue.ts
breadchris: ill add a TODO in the code that if someone needs to process other aws data sources in this queue that they will need to check the event source.

PR #623, opened 2022-05-13
Mockable github webhook

breadchris:
Adds environment flag to disable the webhook for the backend.

Fixes insecure clone url git:// -> https://

Updates grype and syft dependencies to latest.


ajvpot (approved):

lunatrace/bsl/backend/src/config.ts
```diff
@@ -67,9 +68,17 @@ export function getAwsConfig(): AwsConfig {
 }
 
 export function getWebhookConfig(): WebhookConfig {
+  const shouldDisableWebhookQueue = checkEnvVar('DISABLE_WEBHOOK_QUEUE', 'false');
+
+  const disableWebhookQueue = validateBooleanString(shouldDisableWebhookQueue);
+  if (disableWebhookQueue.error) {
```
factoidforrest, line 74: Nice check. Tiny nit and doesn't need fixing but I can see the golang code leaking through.  For JS I would probably just have that function itself throw and return a boolean instead of a wrapped boolean.

lunatrace/bsl/backend/src/workers/handler.ts
```diff
@@ -184,21 +184,35 @@ function wrapProcessRepositoryJob(
...
-      handlerFunc: wrapProcessRepositoryJob(generateSnapshotForRepository),
+      // eslint-disable-next-line @typescript-eslint/require-await
+      handlerFunc: async () => {
+        return wrapProcessRepositoryJob(generateSnapshotForRepository);
+      },
     },
     'process-sbom': {
-      handlerFunc: wrapProcessS3EventQueueJob(handleScanSbom),
+      // eslint-disable-next-line @typescript-eslint/require-await
+      handlerFunc: async () => {
```
factoidforrest, line 203: Im not sure why we had to wrap all of these in async.  Maybe in an effort to get async errors to throw?  I have a feeling we could DRY this up by doing what youre doing here in the `readDataFromQueue` function that is taking this callback.  

Again, tiny nit, doesnt need fixing.

factoidforrest (approved):
sweet

PR #631, opened 2022-05-14
Turn off pr comments

factoidforrest:
Haven't tested this on the backend yet but its so simple I think it will just go. Please don't merge until I get a change to test it though. Very happy with this react UX, personally.
shot_048


lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -0,0 +1,60 @@
+ALTER TABLE public.settings ADD COLUMN pr_feedback_enabled boolean DEFAULT true NOT NULL;
+ALTER TABLE public.settings DROP COLUMN is_org_settings;
+ALTER TABLE public.projects ADD CONSTRAINT project_settings_fkey FOREIGN KEY (settings_id) REFERENCES public.settings;
+ALTER TABLE public.organizations ADD CONSTRAINT organization_settings_fkey FOREIGN KEY (settings_id) REFERENCES public.settings;
+
+--  get caught up on the missing settings rows
+DO $$
```
factoidforrest, line 7: this is uh..this part.. well.

lunatrace/bsl/frontend/src/pages/project/Main.tsx
```diff
@@ -90,8 +108,8 @@ export const ProjectMain: React.FunctionComponent = (_props) => {
         return <Builds project={p} />;
       case 'secrets':
         return <SecretsMain project={p} />;
-      //{/*case 'settings':*/}
-      //{/*  return <ProjectSettingsMain project={p} />;*/}
+      case 'settings':
```
freeqaz, line 111: Super pro, nice 😎

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -0,0 +1,60 @@
...
+    if (NEW.settings_id IS NULL) then
+        INSERT INTO public.settings DEFAULT VALUES RETURNING id INTO NEW.settings_id;
+    end if;
+    RETURN NEW;
+END;
+$BODY$
+    LANGUAGE plpgsql VOLATILE;
+
+
+CREATE TRIGGER add_project_settings
```
freeqaz, line 50: I wonder if Triggers work with Aurora. They probably do. We'll find out if the migration fails 🤷‍♂️

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
ajvpot: could it be easier to lazily create settings whenever they are set to a non default value?

freeqaz (approved):
Looks pretty dope. Thanks for shipping this fast!

ajvpot (approved):
ux for this looks great!

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
factoidforrest: technically yes and it did occur to me, but such things are much easier when you control your own backend.  We..have hasura.  This works, though. It's done and it should keep things totally locked together.

ajvpot: normally I'd agree but I foresee a potential issue with the current implementation:
If the settings row doesn't exist for whatever reason (trigger is broken, trigger got removed, etc), that's a [hard error](https://github.com/lunasec-io/lunasec/pull/631/files#diff-7ba204850c5380d2c5ad51d91f3904a224746af03038ae3a40af37af15080f5aR81) which would currently be silent/swallowed by nonexistent logging alerts. It would not be clear about how this row comes to exist while trying to debug that failure.

I think going back to the default behavior if the row doesn't exist makes more sense and results in less complexity. Ultimately your call, I don't want to slow us down but I think it's important to be careful about how we design database schema.

A note on defaults, I think it would also make sense to make the default values for settings all the zero value of the type. For instance, the setting would be "disablePrComment," type boolean, default value is false which is true in all languages. Then you don't need to hardcode default values.

factoidforrest: Yeah still on the opposite side on this one. This type of in-database trigger is literally atomic when projects and orgs get inserted. Given this code, it will be impossible for a project or org to ever exist without settings.  

Defaulting to an empty and only inserting on change is much trickier.  Because this is consumed in multiple places such as the frontend, backend, queue, potentially CLI.... Now they all need to implement that behavior and be responsible for defaulting to the same defaults, and know how to create the whole settings row when needed.  Yeah, if it was one consumer I'd see it, but with a distributed system and multiple languages and all the complexity we already have, I'd much rather just have this sorted out by the database in one place, and then everyone else can be naive.

As much as I hate having this much logic in SQL, this is what we bought when we bought into hasura.

PR #635, opened 2022-05-16
paginate project builds kind of

factoidforrest:
Actually dont merge yet i need to make an index for that aggregate or its going to be slow as heck


lunatrace/bsl/frontend/src/api/graphql/getProject.graphql
```diff
@@ -28,10 +28,10 @@ query GetProject($project_id: uuid!) {
             created_at
             last_used
             created_by_user {
-              traits
+                traits
             }
         }
-        builds(order_by: {created_at: desc}) {
+        builds(order_by: {created_at: desc}, limit: $build_limit) {
```
ajvpot, line 34: It would be cool to do offset/limit pagination here but I can see how that would get really difficult w/ RTK. People probably won't scroll down that far anyway.

ajvpot (approved):

PR #637, opened 2022-05-17
License scanning command line tool

ajvpot:
Creates a CLI tool which is the entrypoint for license anaylsis. Can be run on local files.

Also:
* A dummy package is inserted to database
* Schema works
* Codegen works


freeqaz (approved):

lunatrace/bsl/license-worker/internal/pkg/scanner/test/package/cjs/react-jsx-dev-runtime.profiling.min.js
```diff
@@ -0,0 +1,39 @@
+/*
```
freeqaz, line 1: Do these files need to get nuked?

ajvpot: These are test files, I could make them shorter but I think they're good for testing if header scanning works.

PR #639, opened 2022-05-17
settings are upserted when they are created

breadchris:
When a project's settings are changed, either a new row is created, or an existing row is updated.


lunatrace/bsl/hasura/migrations/lunatrace/1652730865412_package_schema/down.sql
```diff
@@ -1 +1 @@
-DROP SCHEMA IF EXISTS "package";
+DROP SCHEMA IF EXISTS "package" CASCADE;
```
breadchris, line 1: these changes are from Alex's broken down migration for packages

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -1,60 +1,14 @@
...
-        FOR organization IN SELECT * FROM public.organizations WHERE settings_id IS NULL
-            LOOP
-                WITH setting AS (INSERT INTO public.settings DEFAULT VALUES RETURNING id)
-                UPDATE public.organizations
-                SET settings_id=setting.id
-                FROM setting
-                WHERE public.organizations.id = organization.id;
-            END LOOP;
-    END; $$;
+ALTER TABLE public.settings ADD COLUMN "project_id" uuid null;
```
breadchris, line 7: project reference is now a back reference instead of a forward reference. this lets us upsert into the settings table from a project page.

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -1,60 +1,14 @@
...
-FOR project IN SELECT * FROM public.projects WHERE settings_id IS NULL
-LOOP
-    WITH setting AS (INSERT INTO public.settings DEFAULT VALUES RETURNING id)
-    UPDATE public.projects
-    SET settings_id=setting.id
-    FROM setting
-    WHERE public.projects.id = project.id;
-END LOOP;
-END; $$;
+ALTER TABLE public.settings ADD COLUMN pr_feedback_disabled boolean DEFAULT false;
```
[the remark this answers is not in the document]
factoidforrest: I don't think this follows the "nullable" pattern that we talked about on slack.  The part of having it default to null that I like  is that we can detect when a user never set the column, that way we can change the default if necessary or override unset settings from a higher-order group such as at the organization level.  

Treating the entire settings row as nullable isn't the right abstraction to achieve that, since it will only tell us if ANY setting has been changed, not if a specific setting has been set. To achieve what we talked about, I would make the column `boolean DEFAULT NULL;` if you want to be explicit. 

Having the entire row be nullable is just an inconvenient thing we need to code around in our feature code.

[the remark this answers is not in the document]
factoidforrest: I don't think this follows the "nullable" pattern that we talked about on slack.  The part of having it default to null that I like  is that we can detect when a user never set the column, that way we can change the default if necessary or override unset settings from a higher-order group such as at the organization level.  

Treating the entire settings row as nullable isn't the right abstraction to achieve that, since it will only tell us if ANY setting has been changed, not if a specific setting has been set. To achieve what we talked about, I would make the column `boolean DEFAULT NULL;` if you want to be explicit. 

Having the entire row be nullable is just an inconvenient thing we need to code around in our feature code.

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
factoidforrest: The reason it was the other way is because there is a polymorphic relation from <Table> -> settings.  If you want to build a polymorphic relation with the foreign key in the child table you have to do something more complex. 

It's maybe not worth having the polymorphic relationship since the settings aren't exactly 1:1 anyway. The idea was that most of the columns would be overlapping since org level settings are mainly for overriding/bulk setting project settings.  That was the idea. Maybe too-smart-by-half but we've lost that ability here.

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -1,60 +1,14 @@
...
--- make sure any new projects or orgs have a settings table associated
-CREATE OR REPLACE FUNCTION always_create_settings()
-    RETURNS trigger AS
-$BODY$
-BEGIN
-    if (NEW.settings_id IS NULL) then
-        INSERT INTO public.settings DEFAULT VALUES RETURNING id INTO NEW.settings_id;
-    end if;
-    RETURN NEW;
-END;
```
factoidforrest, line 45: We all need to get comfortable with writing SQL, if we are going to keep Hasura.  This is the tech stack we chose.  We can't on one hand say we can fix the issues with hasura by "just write the SQL" and then be too wary of SQL to actually write SQL and trust it to work reliably.  I don't love it either, but this is truly fine, normal database stuff.

lunatrace/bsl/frontend/src/pages/project/settings/PrComments.tsx
```diff
@@ -26,39 +27,45 @@ export const SettingsPrComments: React.FC<SettingsPRCommentsProps> = ({ project
...
   return (
     <>
       <Row>
         <Col md="4">
-          <h2>Pull Request Feedback</h2>
+          <h3>Pull Request Feedback</h3>
         </Col>
         <Col md>
-          <p>Whether LunaTrace will automatically submit feedback your PRs.</p>
+          <p>Disable LunaTrace from submitting feedback to your PRs.</p>
```
factoidforrest, line 49: This is a confusing UX now.  So PR comments are only ON when the switch is OFF? It's backwards.

"Dont make me think"  

I'm perfectly happy to change the underlying data structure to be less semantic if it enables some useful business logic, but there's no reason to have a confusing UX.

factoidforrest (changes requested):
I picked up some issues here.  

* The polymorphic relationship of settings to an arbitrary table is broken
* The UX is backwards from what a user would expect.  We can handle weird logic as devs but we shouldn't pass it on to users. They are going to be clicking around fast, skimming text. It needs to be straightforward, unless theres a good reason to make it confusing. 
* We still don't know if a user has ever changed a specific setting or not. We moved the "nullable" column up to the row level, where the abstraction isnt useful except to say if the user has ever changed ANY setting. Having the row nullable let us get rid of the SQL triggers (which work btw) but moves the complexity into the client code where it is more likely to introduce bugs. 
* This moves the responsibility for knowing what the default behavior is into the feature code. That is ok, and maybe its better if we were using default nullable columns, but its something to be aware of. To maintain best practice as we have more clients calling and using this data, we should probably stick to the "default" value being false, rather than the semantic value that might be a little easier to understand.  


The idea I'm excited about from our discussions that I would like to bring in is the default null columns because it let's us do smarter setting inheritance, but that isn't present in this PR.  This broke the polymorphic relation.

PR #640, opened 2022-05-17
Fix down migrations, set default project schema to PostgreSQL

ajvpot:
Pulls in down migration changes from #639 
Sets default dialect for IntelliJ SQL to Postgres.


factoidforrest (approved):
beauty

factoidforrest (approved):

PR #641, opened 2022-05-17
Settings rework

factoidforrest:
Hopefully solves the issues raised by chris and alex about the sketchy SQL and implements their requested changes around flags defaulting to NULL.  Thanks for Free for the SQL fix.

* Populates settings_id with `DEFAULT` instead of with trigger.  More referentially secure and allows us to make the column `NOT NULL` without hasura messing up, because it will see the default and omit it from required fields.
* Sets the settings flag to NULL by default so that we can see if a setting has been changed or not.  Flips the boolean around to standardize on `false` being the stock value, so that NULL and FALSE values both resolve as falsey in client code, allowing us to do things like: `if (pr_feedback_disabled) {....` This will enable us to merge settings more naively to create inheritance.  The null essentially turns the binary into a trinary and gives us a third value to capture that with :)

Oh and I tested it, it works fine. 

THIS EDITS THE MIGRATIONS IN PLACE ON MASTER which is a no-no but since this isn't deployed we figured it was ok.  The down migrations are equally overwritten so every developer must migrate down below the migration in question 
`hasura migrate apply --down 3`, take the change, and then migrate up again. Sorry about that. You may need to `git checkout` an earlier commit to do the down if you find yourself in a broken DB state on master.  Please let me know if anything goes haywire.


lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/down.sql
```diff
@@ -1,10 +1,25 @@
...
 
-ALTER TABLE public.projects DROP CONSTRAINT project_settings_fkey;
-ALTER TABLE public.organizations DROP CONSTRAINT organization_settings_fkey;
+ALTER TABLE public.projects DROP COLUMN settings_id;
+ALTER TABLE public.projects ADD COLUMN settings_id uuid REFERENCES public.settings;
 
-ALTER TABLE public.settings DROP COLUMN pr_feedback_enabled;
-ALTER TABLE public.settings ADD COLUMN is_org_settings boolean;
+ALTER TABLE public.organizations DROP COLUMN settings_id;
+ALTER TABLE public.organizations ADD COLUMN settings_id uuid REFERENCES public.settings;
```
ajvpot, line 6: Do we need to add tables to support organizational inheritance if it is not implemented? imo we should keep the schema as simple as possible until we have a plan for what we are going to do.

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
```diff
@@ -1,60 +1,19 @@
...
+CREATE OR REPLACE FUNCTION public.insert_default_settings()
+    RETURNS uuid AS
 $BODY$
+DECLARE new_id uuid;
 BEGIN
-    if (NEW.settings_id IS NULL) then
-        INSERT INTO public.settings DEFAULT VALUES RETURNING id INTO NEW.settings_id;
-    end if;
-    RETURN NEW;
+    INSERT INTO public.settings DEFAULT VALUES RETURNING id INTO new_id;
```
ajvpot, line 8: This may not solve the issue with nullability. If we're trying to implement inheritance, inserting the default values for every project will override org-wide settings in all cases.

freeqaz: What about if the columns in the settings table are nullable? Then there can be a fallback to the organization settings.

Regardless, given that we don't actually have inheritance implemented yet, I vote that we punt this and just pick the easy solution that is future proof. (Even if that means only _new_ projects get the inheritance because the current values are non-nullable, that's fine.)

factoidforrest: > that is precicely how it is implemented as you see here

ajvpot (approved):

lunatrace/bsl/hasura/migrations/lunatrace/1652400983913_add-pr-comment-disable/up.sql
factoidforrest: >  If we're trying to implement inheritance, inserting the default values for every project will override org-wide settings in all cases.
The defaults ARE null values so I think that fixes it

Issue #688, opened 2022-05-27
Make github install route blocking


breadchris:
The github install route was refactored to be async which causes an installation to return instantly. This causes stale data to be served back to the user.

Ideally we poll for completion on the action, but to get this done for now we should add back the blocking behavior to this handler so that we can ship something quickly.

PR #693, opened 2022-05-31
Report status during install

breadchris:
When an install is performed, there is a possibility that a user is sent back to lunatrace after going to Github, but the projects have not been populated yet. If this is the case, then it will appear as if the app is broken.

Waiting for the install event to at the very least create the organization/projects in the database will improve UX during install.

### DIFF
--- a/lunatrace/bsl/backend/src/github/api/generated.ts
+++ b/lunatrace/bsl/backend/src/github/api/generated.ts
@@ -24019,7 +24019,7 @@ export type GetMembersForOrganizationQueryVariables = Exact<{
 }>;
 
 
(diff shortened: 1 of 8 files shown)


breadchris:
#688

factoidforrest (approved):
Went ahead and fixed that query.  You can query relations in a where clause, and also just query for counts directly.  No reason to query data you dont actually want, such as organizations.  Makes the code confusing because I assume you WANT organizations for something, but you didn't.  


I'll be more comfortable when this checks that the installed repos are all present, and I don't think that would be much harder, but this is definitely  better than we had it!

Nice fix.

PR #715, opened 2022-06-08
installation waits for specific IDs

factoidforrest:
I tested this manually in a few scenarios and it seems to work fine.  It definitely waits until installed.  I didn't simulate the error case... could write some tests if we feel sketch.


lunatrace/bsl/backend/src/express-routes/github-routes.ts
```diff
@@ -18,5 +18,5 @@ import { githubLogin } from '../github/login';
 
 export const githubApiRouter = express.Router();
 
-githubApiRouter.get('/github/install', githubInstall);
+githubApiRouter.get('/github/install', express.json(), githubInstall);
```
breadchris, line 21: we could just apply this as middleware for all of `githubApiRouter` all requests coming in via github will be json

lunatrace/bsl/backend/src/github/install.ts
```diff
@@ -113,10 +122,10 @@ export async function githubInstall(req: Request, res: Response): Promise<void>
     return;
   }
 
-  const installError = await waitForGithubInstall(installationId);
+  const installError = await waitForGithubInstall(installationId, installationAuthToken.res);
 
   if (installError.error) {
-    log.error('unable to verify that Github App was installed successfully', {
+    log.error(installError.msg, {
```
breadchris, line 128: we should keep the specific log message that was removed and put the key `error: installError.msg` in the object of this logging call. This will help us track down where an error is coming from since we don't have line numbers when we log.

breadchris (approved):
a couple of nits, but otherwise is good

PR #719, opened 2022-06-10
Search ignored findings

breadchris:
screenshot-localhost_4455-2022 06 09-23_11_57

screenshot-localhost_4455-2022 06 09-23_13_10

Allows someone to search findings for vulns that have been ignored.

- If a package has all of its findings ignored, then "Ignored" will appear next to the package name
- If a package only has some of its findings ignored, then the ignored findings will have the reason they have been ignored.


lunatrace/bsl/frontend/src/components/finding/VulnerabilityTableItem.tsx
```diff
@@ -11,15 +11,16 @@
  * limitations under the License.
```
factoidforrest, line 11: The finding component is in the "project" directory because it is a single use component that only lives in this page.  If you think the page is unclear, we could break the page out of project and have it in its own "snapshot/build" folder, since it's a little deep.  Components are for reusable components like buttons, headers, utility components, etc.

Finding will probably stay attached to the snapshot since this a very snapshot-centric view.

lunatrace/bsl/frontend/src/components/finding/VulnerablePackageCardBody.tsx
```diff
@@ -0,0 +1,208 @@
+/*
```
factoidforrest, line 1: same thing.  This belongs in pages

lunatrace/bsl/frontend/src/components/finding/VulnerablePackageList.tsx
```diff
@@ -11,25 +11,30 @@
...
-import { groupByPackage, severityOrder } from '@lunatrace/lunatrace-common';
-import React, { useState } from 'react';
+import { groupByPackage, severityOrder } from '@lunatrace/lunatrace-common/build/main';
+import React, { ChangeEvent, useState } from 'react';
 import { Col, Container, Dropdown, Row } from 'react-bootstrap';
 
 import { VulnerablePackageItem } from './VulnerablePackageItem';
 import { Finding } from './types';
 
+export interface QuickViewProps {
```
factoidforrest, line 21: nice and i almost did this when i wrote it.  Since we use this all over, why not move it into the types file.

lunatrace/bsl/frontend/src/utils/string-utils.ts
```diff
@@ -21,3 +21,7 @@ export function truncateString(str: string, maxLength: number) {
   }
   return str;
 }
+
+export function maybePluralize(count: number, noun: string, includeCount = false, suffix = 's') {
```
factoidforrest, line 25: pluralizeIfMultiple

factoidforrest (approved):

ajvpot (approved):

PR #732, opened 2022-06-14
Do not show all orgs if user is admin

breadchris:
All organizations were being shown in the sidebar if the user was admin. This will only show the organizations that the user is a user of. Using the admin page, more organizations can be joined.

### DIFF
--- a/lunatrace/bsl/frontend/src/contexts/SidebarContext.tsx
+++ b/lunatrace/bsl/frontend/src/contexts/SidebarContext.tsx
@@ -11,12 +11,27 @@
  * limitations under the License.
  *
  */
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
+import api from '../api';
+import { GetSidebarInfoQuery } from '../api/generated';
 import { SIDEBAR_BEHAVIOR, SIDEBAR_POSITION } from '../constants';
+import useAppSelector from '../hooks/useAppSelector';
 import useSettingsState from '../hooks/useSettingsState';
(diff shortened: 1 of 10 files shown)


lunatrace/bsl/frontend/src/contexts/SidebarContext.tsx
```diff
@@ -11,12 +11,27 @@
...
-const initialState = {
+interface SidebarContextType {
+  isOpen: boolean;
+  position: string;
+  behavior: string;
+  setIsOpen: (_a: boolean) => void;
+  setPosition: (_a: string) => void;
+  setBehavior: (_a: string) => void;
+  sidebarData: GetSidebarInfoQuery | undefined;
+  sidebarIsLoaded: boolean;
```
factoidforrest, line 31: So there isn't actually any reason to store the data in this context. I think actually the move is to rename the query to "getAllPageLoadData" or something.  This started as information needed by the sidebar but it ended up being information needed all over the app. With rtk it's more idiomatic and overall a better idea to simply call the same endpoint from all over the app. That way you can pull out any additional information you might need in that component.  


Really this is my fault for not changing the name before you got to it.

factoidforrest (changes requested):

lunatrace/bsl/frontend/src/contexts/SidebarContext.tsx
factoidforrest: Rtk is behind the scenes automatically using the redux store.  Let it do the work

factoidforrest (approved):

PR #740, opened 2022-06-16
default branch builds happen on PR merge or direct push

factoidforrest:
Seems to work fine. "push" actually fires on merges, so no second listener needed.

 Relatively simple. Super nice now that our webhook code is so good, its very easy to hook up new webhooks like this.  Good job us.

Looks like more changes than it is because it was time to break the handlers into their own files, so I did that as well.


breadchris (approved):
looks good. breaking the webhook handlers out into different files was a good idea.

PR #741, opened 2022-06-17
resync all repos whenever any are updated

factoidforrest:
Fire the repository resync when we get a repository.edit webhook.  Did this particularly because it keeps the default_branch in sync, and it will keep anything else in sync also which is nice. 

It's an absolutely heinous overfetch but..it was certainly quick to write haha.  And realistically it's probably fine, its just a little speed-debt.


breadchris (approved):
good meme

PR #772, opened 2022-07-11
Better vulnerability data

factoidforrest:


lunatrace/bsl/ingest-worker/pkg/vulnerability/advisory/osv.go
```diff
@@ -334,13 +350,64 @@ func mapAffectedElement(element schema.OsvSchemaAffectedElem) *gql.Vulnerability
...
+	queryBody := &gql.Vulnerability_affected_insert_input{
 		Affected_range_events: rangeEvents,
 		Affected_versions:     mapAffectedVersion(element.Versions),
 		Database_specific:     util.Ptr(databaseSpecific),
 		Ecosystem_specific:    util.Ptr(ecosystemSpecific),
 		Package:               affectedPackage,
 	}
+
+	if len(element.Ranges) != 0 {
+		queryBody.Ranges = buildRangeInsert(element.Ranges)
```
factoidforrest, line 362: This is the part im least sure about.  It was throwing errors on insert if this key was in the struct so I decided to just not add it unless there are any ranges to work with. Im sure theres a better way, was just stumbling through

lunatrace/bsl/ingest-worker/pkg/vulnerability/advisory/osv.go
```diff
@@ -334,13 +350,64 @@ func mapAffectedElement(element schema.OsvSchemaAffectedElem) *gql.Vulnerability
...
+					slug += "introduced:" + sanitizedVersionString
+				}
+				if eventPropName == "fixed" {
+					parsedRange.Fixed = &sanitizedVersionString
+					slug += "fixed:" + sanitizedVersionString
+				}
+			}
+		}
+
+		if existingRangeSlugs[slug] != true {
```
factoidforrest, line 405: If you insert a duplicate that violates the conflict in the SAME query, postgres will throw, so we have to pre check.  Maybe theres a better way

breadchris: this seems like the right thing to do here for now

Issue #766, opened 2022-07-05
Upsert query for inserting package relationships into the db


breadchris:
The relationships between packages, as described in the lock file, are helpful when attempting to do an automatic version bump. Using this project https://github.com/snyk/nodejs-lockfile-parser to collect these relationships, we will use the output and push it to the database.

PR #773, opened 2022-07-12
Collect lockfile contents

breadchris:
The transitive dependencies for packages are currently not being collected. Using the library https://github.com/snyk/nodejs-lockfile-parser will give lunatrace some much needed context to where a package comes from. Determinations during reporting can be made to help rank what vulnerabilities should be surfaced or not.


breadchris:
#766

lunatrace/bsl/backend/src/utils/fs.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * https://github.com/lunasec-io/lunasec/blob/master/licenses/BSL-LunaTrace.txt
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import fs, { Dirent } from 'fs';
+
+export function* walk(path: string, filter: (path: string, name: string) => boolean): IterableIterator<string> {
```
factoidforrest, line 16: using a generator function and yield for this seems really peculiar.  I can't figure out the reasoning, especially when combined with using readdirSync which has performance issues since that blocks the node thread.  Maybe this is why you and I disagreed about how fast node could recurse a directory :p.  This will only be able to read and recurse one directory at a time, where as doing it async will be more like log(n) or at least do a lot more in parallel.

Rather than use a generator, I would just make a normal async function that returns an array of path matches.  Inside that you can have an inner recursive async function that does the walking.  

In addition to generators, you almost never see `for of` in the wild.  .forEach is much more common

lunatrace/bsl/backend/src/utils/io.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { spawn, SpawnOptionsWithoutStdio } from 'child_process';
+
+import BufferList from 'bl';
+
+export const asyncSpawn = (command: string, args?: ReadonlyArray<string>, options?: SpawnOptionsWithoutStdio) => {
```
factoidforrest, line 18: this is fantastic

lunatrace/bsl/backend/src/snapshot/package-tree.ts
```diff
@@ -0,0 +1,165 @@
...
+  return flattenedDependencies
+    .map((entry): Build_Dependency_Relationship_Insert_Input | null => {
+      const { id, dependency, parentId } = entry;
+      if (!dependency.version || !dependency.name) {
+        return null;
+      }
+
+      return {
+        /*
+         TODO (cthompson) we are generating the ids for the database here. This avoids a roundtrip from the server
```
factoidforrest, line 90: You can delete this comment
> [As per Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier), the number of UUIDs generated to have atleast 1 collision is 2.71 quintillion.
> 
> This is equivalent to generating around 1 billion UUIDs per second for about 85 years. Oh Boy!! that’s a really long time.
> 
> 
Basically, we never worry about this actually happening. 

Another option, to avoid generating these UUIDs on the client out of laziness and concern about bugs/complexity, would be to just build the entire tree in the correct shape for hasura, in memory, and send it up in that shape.  Hasura figures out the relationships and builds the foreign keys like a good ORM should :) I understand that's a bit of work to get it into that structure since the manifest files are flat and things don't really get parsed in order.  Just another option I wanted to mention.

lunatrace/bsl/backend/src/snapshot/package-tree.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  const buildDependencyRelationships = pkgTrees
+    .map((pkgTree) => mapPackageTreeToBuildDependencyRelationships(buildId, pkgTree))
+    .flat();
+
+  log.info('generated package tree for pinned dependencies', {
+    count: buildDependencyRelationships.length,
+  });
+
+  const chunkSize = 5000;
```
factoidforrest, line 143: Do we actually need chunking here?  How can we be sure it wont insert a deeper item in the tree before it inserts a higher item and violate the foreign key constraint? Are they in order? Can we just try doing them flat until it breaks? Are we hitting some built in limit in hasura?  They have some security stuff around this in their enterprise stuff but its really hard to figure out whats going on in our OSS version. https://hasura.io/docs/latest/graphql/core/databases/postgres/mutations/upsert/#pg-nested-upsert-caveats

lunatrace/bsl/backend/src/snapshot/package-tree.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    const chunk = buildDependencyRelationships.slice(i, i + chunkSize);
+    log.info('inserting build dependency relationships', {
+      chunkSize: chunk.length,
+    });
+
+    const resp = await hasura.InsertBuildDependencyRelationships({
+      objects: chunk,
+      on_conflict: {
+        constraint: Build_Dependency_Relationship_Constraint.BuildDependencyRelationshipPkey,
```
factoidforrest, line 157: This in particular I would leave out because now if there ever IS a collision, it will just fail silently, change the pkey and not anything else, and maybe even try to break some other foreign keys creating a weird error.  

Do we actually need this for anything? Isn't it better to outright fail?

lunatrace/bsl/backend/src/fixtures/large-repo/package.json
```diff
@@ -0,0 +1,122 @@
+{
+  "author": "",
```
factoidforrest, line 2: this was missing the lockfile because the gitignore ignored it, causing tests to fail.

PR #791, opened 2022-07-21
fix guides and some ugly vuln stuff

factoidforrest:
guide upserting was broken.  Need to migrate prod before merging this, i can do it.


lunatrace/bsl/hasura/migrations/lunatrace/1658439316277_fix-guides-fkey/up.sql
```diff
@@ -0,0 +1,5 @@
+--  this just wipes out the whole table, it's fine, we will just upsert next time anyway
+DELETE FROM public.guide_vulnerabilities;
+ALTER TABLE public.guide_vulnerabilities DROP COLUMN vulnerability_id;
+ALTER TABLE public.guide_vulnerabilities ADD COLUMN vulnerability_id uuid references vulnerability.vulnerability (id) NOT NULL;
```
factoidforrest, line 4: This is the main change of this PR

guides/LUNAGUIDE-20220422-1-LOG4SHELL/metadata.yaml
```diff
@@ -1,7 +1,7 @@
 schemaVersion: 1
 cves:
-  - CVE-2021-44228
-  - CVE-2021-45046
+  - GHSA-jfh8-c2jp-5v3q
```
ajvpot, line 3: These are not CVEs?

ajvpot (changes requested):
Looks good except metadata changes, if the upsert job can't handle cves in this field and resolve them we should probably make it.

ajvpot (approved):

breadchris (approved):
aside from the `cves` section on the guide that alex pointed out, looks good

PR #795, opened 2022-07-22
rewrite the mailchimp form thing

ajvpot:


docs/src/components/Mailchimp.jsx
```diff
@@ -0,0 +1,70 @@
+/* eslint react/prop-types: 0 */
+import jsonp from 'jsonp';
+import React from 'react';
+
+export default class Mailchimp extends React.Component {
```
factoidforrest, line 5: It's maybe a bit strange to use class components when function components have trounced them so heavily in the last year or so.  Doesn't really matter, just the way things are going.

factoidforrest (approved):

docs/src/components/Mailchimp.jsx
ajvpot: I was having some issues figuring out how to break function components into multiple functions without passing around state and their setters. We should chat about this

PR #846, opened 2022-08-26
Inserter for build dependency relationships using merkle tree

freeqaz:
I still need to test this, but this is built using SQL only and I've done a basic first pass at optimizing it for speed. There is _way_ more we could do, but this should help unblock us for now.


lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -11,180 +11,291 @@
...
+             VALUES ($1, $2)
+             ON CONFLICT DO NOTHING
+             RETURNING id`,
+          [buildId, pkgGraph.lockFilePath]
+        );
+
+        log.info(`Inserted manifest ${manifestId.id} for ${buildId}`);
+
+        const dependencyColumns = new pgp.helpers.ColumnSet(['manifest_id', 'labels', 'manifest_dependency_node_id'], {
+          table: 'manifest_dependency',
```
freeqaz, line 278: This line is using the helper

ajvpot (approved):

lunatrace/bsl/backend/src/tests/node-package-tree.test.ts
```diff
@@ -37,61 +78,72 @@ function testAllTreeTypes() {
...
+      //
+      //   (objectsInserted as Array<Build_Dependency_Relationship_Insert_Input>).forEach((dep) => {
+      //     if (!dep.range) {
+      //       console.log('encountered dep with null range: ', util.inspect(dep, { depth: Infinity }));
+      //     }
+      //     expect(dep.range).toBeTruthy();
+      //   });
+      // });
 
-      describe('root dependency counts', () => {
```
factoidforrest, line 64: Are you still checking these values are correct? Is the idea of the test alive in some other check or test somewhere?  Same for any other deleted test here. Are we still doing the tests somewhere in a different way but with the same intent?

factoidforrest (approved):
nice, pretty straight forward. JW if those deleted tests are replaced with anything

ajvpot (changes requested):
Looks like some extraneous changes snuck in to `lunatrace/cli/cmd/log4shell/Log4ShellHotpatch.class`

PR #847, opened 2022-08-30
Faster merkle tree

freeqaz:


lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -176,9 +170,12 @@ async function insertPackageGraphsIntoDatabase(buildId: string, pkgGraphs: Colle
 
   log.info(`Found ${uniqueNodeRootIds.size} unique root dependency hashes`);
 
-  const currentlyKnownRootIds = await db.manyOrNone('SELECT id FROM manifest_dependency_node WHERE id IN (SELECT $1)', [
-    pgp.as.array(Array.from(uniqueNodeRootIds)),
-  ]);
+  const currentlyKnownRootIds = await db.manyOrNone(
+    `SELECT id FROM manifest_dependency_node WHERE id IN (${pgp.as
```
factoidforrest, line 174: is there a possible race condition here? I guess not..if two inserts happen at the same time, it will just be a bit of wasted compute.  So it's really better to query here and do this in application code than write it into the SQL query huh? definitely simpler, anyway

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -202,20 +199,22 @@ async function insertPackageGraphsIntoDatabase(buildId: string, pkgGraphs: Colle
 
   log.info(`Total ${dependencyNodeMap.size} unique transitive dependency hashes`);
 
-  const currentlyKnownTransitiveDependencyIds = await db.manyOrNone<string>(
-    'SELECT id FROM manifest_dependency_node WHERE id IN (SELECT $1)',
-    [pgp.as.array(Array.from(dependencyNodeMap.keys()))]
-  );
+  const currentKnownQuery = `SELECT id FROM manifest_dependency_node WHERE id IN (${pgp.as
+    .array(Array.from(dependencyNodeMap.keys()))
```
factoidforrest, line 203: Almost the exact same code as above, should we DRY it?

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
factoidforrest: I guess that makes sense actually,  you're inserting the data flat so hasura doesn't know about the tree. We have to delete the whole sections of the tree while we have the whole thing here in tree shape.

factoidforrest (approved):
Neato, i get it, no changes requested. I look forward to seeing this working on the frontend, lmk when you guys try it out!

PR #849, opened 2022-08-30
No longer stringify context keys in JSON Logger transport

freeqaz:
This makes logging in DataDog go from a "flat" stringified JSON object into a format that we can introspect to write alerts for.

I'm not exactly sure why this was implemented the way it was before, but from our testing this now works better for what we need (debugging production).


factoidforrest (approved):
Aren't you never ever supposed to use object as a type? You should use record<string|symbol|number,unknown>

ajvpot (approved):

PR #860, opened 2022-08-31
More logging to test DB connection

freeqaz:
Working theory is that the new DB needs some firewall rules updated. This will let us test that theory.


factoidforrest (commented):
noice

Issue #810, opened 2022-08-03
Project Chooser Install Flow


factoidforrest:
user should have to manually opt-in their github projects, rather than us eagerly adding all of them from github.

PR #865, opened 2022-09-01
Opt in install

factoidforrest:
Fixes https://github.com/lunasec-io/lunasec/issues/810 

**WARNING**: We must remember to change the github redirect post-install url on the github app through the github management GUI when this deploys, or the UX will break. 

A new install flow where users must opt-in their repos from github by clicking each one. Should really reduce the number of garbage repos that we have to deal with. Also its all synchronous so errors surface immediately, things block, etc. Great for UX. No more waiting for background jobs from webhooks.  

Other changes in this PR: 

- Changed the way that repository webhooks with information about repository changes are processed (ex: repository.edited webhook). Any webhooks with new information are processed directly and inserted into hasura. Before, we were refetching everything in the installation and upserting it all. Now, it just takes the new data and updates it, that's it.
- Made member_added webhook 
- DRYd up the github settings config which i noticed had just been copy-pasted. Also it makes more logical sense now because all github settings are grouped together

Future steps: this gives us a good place to do price differentiation if we want to by limiting the total number of projects allowed to be imported.  It could also consider if projects are public vs private and limit only private ones, and so on. 

shot_062


lunatrace/bsl/backend/src/github/actions/get-repos-from-installation.ts
```diff
@@ -13,54 +13,36 @@
...
-    }
-
-    log.info(`[installId: ${installationId}] Detected paginated GitHub repositories`);
+    return { newItems: response.data.repositories, total: response.data.total_count };
   }
 
-  log.error(`[installId: ${installationId}] Hit max iterations for listing repos for installation`);
-
-  throw new Error('Hit max iterations for listing repos for installation: ' + installationId.toString(10));
+  return walkPagination(iLog, callGithub);
```
factoidforrest, line 40: made the pagination walker its own utility, that takes a callback. Thats what all this is. Just DRYing it up since we now need to read pagination in multiple places

lunatrace/bsl/frontend/src/pages/homepage/cards/SetupWizard.tsx
```diff
@@ -81,9 +81,15 @@ export const SetupWizard: React.FunctionComponent<WizardProps> = ({ compact }) =
...
             <Col md={compact ? '12' : '6'} className="d-grid gap-2">
-              <Button variant={hasProjects ? 'light' : 'primary'} size="lg" href={GithubAppUrl}>
-                <BsGithub className="mb-1 me-1" /> {hasProjects ? 'Add more projects' : 'Connect to GitHub'}
-              </Button>
+              {hasProjects ? (
+                <Button variant="light" size="lg" onClick={() => navigate('/project/import')}>
+                  <BsGithub className="mb-1 me-1" /> {'Add more projects'}
+                </Button>
+              ) : (
+                <Button variant="primary" size="lg" href={GithubAppUrl}>
```
factoidforrest, line 89: send new users straight to github since they dont have any projects to manage anyway

lunatrace/bsl/frontend/src/pages/project/settings/PrFeedback.tsx
```diff
@@ -53,18 +56,32 @@ export const SettingsPrComments: React.FC<SettingsPRCommentsProps> = ({ project
...
-            id="pr-reviews-switch"
-            label={checked ? 'PR Comments Enabled' : 'PR Comments Disabled'}
-            checked={checked}
-            disabled={isLoading}
-            onChange={(e) => {
-              setChecked(!checked);
-              void updateSettings({ id: project.settings_id, settings: { pr_feedback_disabled: !e.target.checked } });
-            }}
-          />
+        <Form className="d-flex align-items-center flex-column">
```
factoidforrest, line 59: Oh and I merged this settings UI from these two components because it wasn't DRY and it didn't make too much sense how it was laid out in the UI.  Both the settings really belong in the same section under github feedback, so now they are

breadchris (approved):

PR #871, opened 2022-09-06
Kubernetes CDK stack

ajvpot:
Creates a CDK stack that deploys a kube cluster in the new account.

### DIFF
--- a/lunatrace/bsl/backend-cdk2/.gitignore
+++ b/lunatrace/bsl/backend-cdk2/.gitignore
@@ -1,0 +1,5 @@
+node_modules
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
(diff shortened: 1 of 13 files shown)


freeqaz (changes requested):
I don't think there is any reason to make a new folder for this. We should just add this as a new "Stack" to the existing CDK folder. We can independently deploy this EKS stuff from the existing stuff too. (Just make a new JSON file for stack-inputs instead of a new folder?)

ajvpot:
> I don't think there is any reason to make a new folder for this. We should just add this as a new "Stack" to the existing CDK folder. We can independently deploy this EKS stuff from the existing stuff too. (Just make a new JSON file for stack-inputs instead of a new folder?)

I wanted the new package versions, we should be good to merge this folder into the other one after you land your updates.

PR #872, opened 2022-09-06
Add health checks to Hasura and bump CDK

freeqaz:
Making this change so that the Hasura waits for the backend before it starts. That'll make it so that the metadata isn't inconsistent when Hasura tries to reach out to the backend for the upstream GraphQL schema.

Also, I hadn't realized that we were using a different version of Hasura in prod than in dev. Now they are synced.


factoidforrest (approved):
Noice

factoidforrest (approved):
Noice. We also should all sync our local CLIs to this.  We've had horrible merge conflicts from the CLIs making slightly different metadata before. I'll check what my current cli version is, I think it's already this.

Issue #879, opened 2022-09-07
Fix slow tree generation for packages on front-end


freeqaz:
Currently we generate the tree from the top down, but we need to generate it from the bottom up.

https://github.com/lunasec-io/lunasec/blob/master/lunatrace/bsl/frontend/src/dependency-tree/builds-dependency-tree.ts#L75

PR #884, opened 2022-09-10
Faster dependency graph generation on the front-end

freeqaz:
There are a few fixes in here:
- Bug fix: Logger now puts parameters handed to it into a field called `params` instead of adding them to the context
- Bug fix: Logger now logs numbers instead of swallowing them
- Bug fix: We are no longer skipping inserting manifests when there are no new dependencies to add (thus breaking the experience)
- Bug fix: We are correctly skipping inserting dependencies even when they exist in the DB (currently we insert them anyway)
- Performance fix: We now crawl the tree from the bottom "up" in order to build the vulnerability chains on the client (much faster)
- Enhancement: Our first front-end unit test :)

Probably more too, but I can't remember now. This is a good chunk of work here and it's quite well tested!


lunatrace/bsl/frontend/src/dependency-tree/builds-dependency-tree.spec.ts
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright by LunaSec (owned by Refinery Labs, Inc)
```
factoidforrest, line 2: jest standard is `.test.ts`

lunatrace/bsl/frontend/src/dependency-tree/builds-dependency-tree.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { DependencyTree } from './builds-dependency-tree';
+import { BuildDependencyPartial } from './types';
+
+it('should generate a dependency tree', () => {
+  const dependencies: Array<BuildDependencyPartial> = [
```
factoidforrest, line 18: this would typically be in a fixtures file with jest, like how the backend is. Ideally that would be a folder within the test folder that jest is configured to ignore and then this would be a `.ts` file in there.  That way you could make it way bigger too, if you want. Not that it needs to be, but it shouldn't be right in the `it`.  I like the effort here to manually create a fake tree with some known outputs, thats a nice fixture

Also, this `it` should be wrapped in a describe. "The dependency tree" "should generate"

factoidforrest (approved):
Noice. some jest nits and concerned about the logger change breaking our datadog stuff, so that needs to get cleared with @ajvpot

factoidforrest:
Oh and i think you broke the logger tests

lunatrace/bsl/logger/src/index.ts
```diff
@@ -154,7 +157,7 @@ export class LunaLogger {
 
       // If the argument is an object, then add its keys to the context object
       if (this.isObject(arg)) {
-        logObject.context = mergeObjectIntoRecord(logObject.context, arg as object);
+        logObject.params = [...(logObject.params || []), arg];
```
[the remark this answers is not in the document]
ajvpot: mergeObjectIntoRecord is probably still needed, it will serialize errors.

lunatrace/bsl/frontend/src/store/store.ts
```diff
@@ -26,7 +26,10 @@ export const store = configureStore({
     alerts: alertsSlice.reducer,
   },
   middleware: (getDefaultMiddleware) => {
-    const middleware = getDefaultMiddleware();
+    const middleware = getDefaultMiddleware({
```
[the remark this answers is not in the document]
freeqaz: It is disabled in production by default. https://redux-toolkit.js.org/api/getdefaultmiddleware/#development

lunatrace/bsl/frontend/src/dependency-tree/builds-dependency-tree.spec.ts
freeqaz: Mostly done. I put a TODO for the fixtures.

lunatrace/bsl/frontend/src/store/store.ts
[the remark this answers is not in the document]
ajvpot: I see. Looks sketchy to disable this but if you absolutely need to it's whatever.

lunatrace/bsl/logger/src/index.ts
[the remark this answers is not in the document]
freeqaz: I made a change that preserves the behavior of both. There are issues with `mergeObjectIntoRecord` where it doesn't log everything correctly. (I was getting bit by them in dev)

With this change, we will log errors into two places as both objects and strings.

ajvpot (approved):

lunatrace/bsl/backend/src/models/scan.ts
```diff
@@ -81,9 +81,9 @@ export async function runLunaTraceScan(sbomStream: Readable): Promise<string> {
       outputBuffers.push(Buffer.from(chunk));
     });
     lunatraceCli.stderr.on('data', (errorChunk) => {
-      // log.warn('lunatrace cli stderr', {
-      //   output: errorChunk.toString(),
-      // });
+      log.warn('lunatrace cli stderr', {
```
breadchris, line 84: was this intentional?

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -255,20 +261,20 @@ async function findCurrentlyKnownDependencies(query: string, manifestIds: string
...
     if (result) {
-      currentlyKnownIds.push(result);
+      for (const { id } of result) {
+        currentlyKnownIds.add(id);
+      }
     }
   }
 
-  const allKnownIds = currentlyKnownIds.flat(1);
+  log.info(`Found ${currentlyKnownIds.size} known dependencies`);
```
breadchris, line 275: the pattern that I prefer for logging is:

```
log.info('Found known dependencies', { count: currentlyKnownIds })
```

it is a lot easier to parse logs when the log message is consistent vs having to read a sentence

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -300,14 +306,14 @@ async function insertPackageGraphsIntoDatabase(buildId: string, pkgGraphs: Colle
     Array.from(uniqueNodeRootIds).sort()
   );
 
-  log.info(`Found ${uniqueNodeRootIds.size} missing root dependency hashes`);
+  log.info(`Found ${currentlyKnownRootIds.size} of ${uniqueNodeRootIds.size} known root dependency hashes`);
```
breadchris, line 309: same as above

lunatrace/bsl/backend/src/models/scan.ts
freeqaz: Yep, I had disabled it because of the annoying spam from 401s in dev.

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
freeqaz: That's a good point

breadchris:
#879

PR #894, opened 2022-09-16
Quieter Hasura logging

freeqaz:
Now that we're logging to DataDog in dev this is going to save us some $ :joy:


ajvpot (changes requested):
I don't think we need to make these changes. We are not spending a significant amount of money on log ingestion, and these logs are useful for debugging production issues.

I'd be for any approach that removes logs from local development if necessary to reduce stdout spam.

PR #895, opened 2022-09-19
Fix two typos in arch install guide

tabell:
Thanks for publishing this guide!


tabell:
It doesn't seem like there's a way to sign the CLA without granting an app access to read my private email addresses, which I'm sorry to say is not worth it to me. Maybe best to close this PR and have someone else on the team make the suggested changes.

breadchris:
Thanks for your suggestions! Your first change in exts -> xts is totally a typo, and for your other change I think I am going to put a disclaimer saying that the disk on your system may vary (since not everyone has the same name on their system).

ajvpot:
@tabell Thank you so much for your submission. You raise a valid point about the CLA process, we're looking into changing it to something simpler so you can get credit for your contribution without providing private info/authing to an app. We want to make the process as simple as possible for contributors. Please stay tuned!

factoidforrest:
@breadchris can we close this?

PR #901, opened 2022-09-26
analysis cache schema

ajvpot:


lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/down.sql
```diff
@@ -0,0 +1,19 @@
+-- Could not auto-generate a down migration.
```
breadchris, line 1:
```
DROP TYPE finding_state;
DROP TABLE pacakge.release_analysis_result;
```

Helpful for dev

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TYPE "finding_state" AS ENUM (
```
breadchris, line 1: I haven't been able to figure out how to update an enum inside of a transaction in psql. Additionally, because of how hasura is with these types, I think we should stick to doing table enums.

ajvpot: @factoidforrest wdyt about this? We previously decided to use native enums for consistency but I think it may be worth cutting over. We can make a migration to switch everything to table enums.

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/down.sql
ajvpot: :+1:

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TYPE finding_state AS ENUM (
+    'vulnerable',
+    'not_vulnerable'
+    );
+
+CREATE TABLE package.release_analysis_result
```
factoidforrest, line 6: I'm not convinced this is a "release_analysis". I don't see a release_id. 

I see a manifest_dependency_edge_id, and since we made that name such a mouthful, seems like we're now avoiding trying to use it.  Maybe this is just a "false_positive_analysis", I'd be fine with that.

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
```diff
@@ -0,0 +1,20 @@
...
+(
+    id                          uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
+    manifest_dependency_edge_id uuid,
+    vulnerability_id            uuid,
+    state                       finding_state,
+    created_at                  timestamptz,
+    source                      TEXT
```
factoidforrest, line 13: what is this "source" column?

lunatrace/bsl/hasura/migrations/lunatrace/1664218140551_alter_table_public_manifest_dependency_edge_add_column_id/down.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."manifest_dependency_edge" drop column "id" cascade
+alter table "public"."manifest_dependency_edge" drop column "id";
```
factoidforrest, line 2: I'm still confused why edges don't just have all the info about their child in them as we had before, it's a one-to-one and it's just a few rows. Now we are getting into complexity here where it turns out that simple join table we had before now has its own identity we want to point to. It's fine but it feels like a consequence of that decision that I still don't quite understand. 

But, you gotta do what you gotta do.

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
```diff
@@ -0,0 +1,20 @@
...
+);
+
+ALTER TABLE package.release_analysis_result
```
factoidforrest, line 16: These foreign keys should just be part of the above definition. 

``` manifest_dependency_edge_id uuid REFERENCES manifest_dependency_edge(id)```

File: lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql [hunk 7]
```diff
@@ -0,0 +1,20 @@
...
+    'vulnerable',
+    'not_vulnerable'
+    );
+
+CREATE TABLE package.release_analysis_result
+(
+    id                          uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
+    manifest_dependency_edge_id uuid,
+    vulnerability_id            uuid,
+    state                       finding_state,
```
factoidforrest, line 11: shouldn't basically ALL of these be NOT NULL?

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
factoidforrest: We know this is a result, but a result of WHAT? What type of analysis was performed? If there are two analysis of the same type, which one was it?  I assume "source" is not an attempt to encapsulate all of that information in a text field. 

I recall from our discussion we decided that analysis types and their definitions would be first class citizens in the database. From that conversation, it was decided that each individual analysis relating to a specific vulnerability (and not just their types) would be represented as a row, to facilitate manual adding/editing of analyses, reviewing of what analyses are possible when viewing a vulnerability, and tracking. 

It follows that for any "auto generated" analyses of the type you envision to populate our early reservoir of analyses, they would be generated at vulnerability insert/upsert time, which recall is a daily job. 

Here is a rough draft of how I think that table would look: 
```sql
CREATE TABLE vulnerability.analysis (
    id                          uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
    auto_generated   bool NOT NULL,
    type                     analysis_type_enum NOT NULL, # excellent application of a table-enum type here, an example row would be "semgrep"
    rule_body jsonb NOT NULL, # having this untyped isnt the end of the world, it needs to be extendable for the different sorts of analysis we want to run. I don't mind a couple more checks to make sure keys arent missing. This can house an entire semgrep rule, no problem. JSONb is truly incredible in PG.
    comment text NULL,
    vulnerability_id uuid NOT NULL,
    package_id uuid NOT NULL,
    semver_range_narrowing text NULL, # we may wish to create rules for package versions that are more specific than what is captured in the vulnerability itself
    skip bool DEFAULT false NOT NULL # not a bad idea to give us a way to short circuit auto generated analyses that might be eroneous without having to hard code it somewhere.
)
    
```

It goes without saying that the time to figure this out is now, not later.

@freeqaz

factoidforrest (changes requested):

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
ajvpot: I would agree about renaming it because "state" is redundant.

Yes, there will be more values. We're eventually hoping to have all the values from the [VEX part of CycloneDX](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.4.proto#L622-L643) and more, so that's why we would want to edit it later.

I think that makes a pretty good case for table enum in this case.

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
ajvpot: We decided to not implement that yet, coming in a later PR. Agree about having a join table later.

factoidforrest: Lets meet about that

factoidforrest: and if you review the document, you can see it's part of v1 here under your items which I think this PR represents
```
Update Database Schema
A place to store the analysis rule for a vulnerability
Change analysis result to point to release_dependency table as the edge
```

ajvpot: I think we should discuss this when Chris is back from the dentist. This diff is one step in delivering v0 of the feature, I think this is the minimum to unblock us and is not the endgame schema for the feature.

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
ajvpot: :+1:

lunatrace/bsl/hasura/migrations/lunatrace/1664218225933_release_analysis_result/up.sql
ajvpot: > what is this "source" column?

Source would be the name of the scanner that generated the result. Open to changing to enum but I left it as text for now.

PR #903, opened 2022-09-28
Tree in backend

factoidforrest:
Backend only, the schema in the frontend is a sample query that could be used to call this resolver, not wired to anything yet.

FIXED The only known broken thing is the severity_name addition to the vulnerability ingester does not seem to be working reliably, many are still null. This may be a more generalized problem with the upserter, or maybe my go code is bad.


lunatrace/bsl/frontend/src/api/graphql/getVulnerableReleasesFromBuild.graphql
```diff
@@ -0,0 +1,49 @@
+query GetVulnerableReleasesFromBuild($build_id: uuid!) {
```
factoidforrest, line 1: sample query for safekeeping/ reference. not wired up yet

lunatrace/bsl/ingest-worker/pkg/vulnerability/advisory/osv.go
```diff
@@ -30,6 +29,26 @@ func isReviewed(rawVuln *schema.OsvSchema) bool {
 	return rawVuln.DatabaseSpecific["github_reviewed"] == true
 }
 
+// ('Critical','High','Medium','Low','Negligible','Unknown')
+func getSeverityName(rawVuln *schema.OsvSchema) string {
```
factoidforrest, line 33: this seems to not work reliably, I think something is janky with the upsert. SOME of them are getting updated, but not all. I'd like to wipe out my DB and see if that fixes the issue, but since we wont have that advantage in production, I'm not sure what the trouble is.

factoidforrest: @ajvpot any ideas why this only upserts a fraction of the vulnerabilities with the proper column? Anything jump out at you?

breadchris: do you have any example vulnerabilities you expect to get upserted, but are not?

PR #906, opened 2022-09-28
Queue handler calls into static analysis

breadchris:


lunatrace/bsl/ingest-worker/pkg/staticanalysis/handler.go
```diff
@@ -0,0 +1,91 @@
...
+			"failed to get package release",
+			util.GraphqlLogContext{
+				Key:   "package release id",
+				Value: queueRecord.PackageReleaseID,
+			},
+		)
+
+	}
+
+	if resp.Package_release_by_pk.Upstream_blob_url != nil {
```
ajvpot, line 85: this looks broken

lunatrace/bsl/ingest-worker/docker/queuehandler.dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.18-alpine AS builder
+
+COPY --from=repo-bootstrap /usr/repo/ /build/
+WORKDIR /build/lunatrace/cli/
+
+# generate the graphql code from the schema
+RUN go generate ./...
```
ajvpot, line 7: This should not be part of the build process. Workflow for go generate is to commit the generated files then build off that.

ajvpot (changes requested):
pls rename the reference in `lunatrace/bsl/ingest-worker/cmd/ingestworker/license/license.go`

PR #909, opened 2022-09-29
move graphql gen out of cli

breadchris:
Go graphql code is now generated with:
```
cd lunatrace/gen && yarn generate
```


lunatrace/gen/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "gen",
+  "private": true,
+  "packageManager": "yarn@3.2.2",
+  "scripts": {
+    "generate:gql:schema": "yarn run gq http://localhost:8080/v1/graphql -H \"X-Hasura-Admin-Secret: myadminsecretkey\" -H \"X-Hasura-Role: service\" --introspect > gql/schema.graphql",
+    "generate:gql:go": "go run github.com/Khan/genqlient genqlient.yaml",
+    "generate:gql": "yarn run generate:gql:schema && yarn run generate:gql:go",
```
factoidforrest, line 8: If this is a golang module designed to sit here and hold the golang query types, cool, but maybe we should name it as such? golang-graphql-gen or something like that? Also does it have this package.json just to run scripts? Surely we could just make a shell script?

I would expect a module named gen to handle ALL of the gen, or else have a more specific name. Is the plan to have it hold all of the generated code? If just the golang code, how about "golang-gen" or something.

I was ok with having the individual TS modules generate their own files, it makes a lot of sense actually since what they generate depends entirely on their own query schemas. I'm not sure I would want them to also end up in this /gen/ folder.

ajvpot: Agree, we should be doing this the standard way which is go:generate.

lunatrace/gen/gql/types/types.go
```diff
@@ -19,62 +19,62 @@ type UUID string
...
-  NPM       = "npm"
-  Packagist = "packagist"
-  CratesIo  = "crates_io"
-  Go        = "go"
-  Hex       = "hex"
-  Maven     = "maven"
-  NuGet     = "nuget"
-  PyPi      = "pypi"
-  RubyGems  = "rubygems"
+	NPM       = "npm"
```
ajvpot, line 22: please dont use tabs

ajvpot (changes requested):

lunatrace/bsl/ingest-worker/pkg/vulnerability/ingester.go
```diff
@@ -15,7 +15,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/Khan/genqlient/graphql"
-	"github.com/lunasec-io/lunasec/lunatrace/cli/gql"
+	gql2 "github.com/lunasec-io/lunasec/lunatrace/gen/gql"
```
ajvpot, line 18: I think the refactor tool added some redundant aliasses

lunatrace/gen/package.json
factoidforrest: So chris did this so he could access "gq" which can fetch the schema and spit it out. I've just remembered, though, that you can do the same thing with the "schema-ast" codegen extension. That's documented absolutely nowhere and even the name is misleading, but that's what it does.  

So if we dont want to use gq we could use the official codegen tool with schema-ast. That's actually exactly what the backend was already doing but it seems @breadchris you deleted that line. Maybe we should bring that back just so my IDE autofill keeps working also, that was nice.

idk, does that make sense?

PR #923, opened 2022-10-08
DataDog RUM Proxy for session debugging

freeqaz:
Please check this for SSRF. I'm pretty sure it's good though.

I implemented this according to this documentation from DataDog: https://docs.datadoghq.com/real_user_monitoring/guide/proxy-rum-data/?tab=latestversion

It leaves a lot to interpretation, so I then tested this against my local box (requires commenting out the Prod-Only RUM check) and it works successfully.

My plan is to just deploy this on an EC2 box, open a port, and then setup an A record in our DNS to point at it. Nothing fancy. No ELB. Just a box serving traffic until the end of time. :P


.yarnrc.yml
```diff
@@ -24,18 +24,35 @@ packageExtensions:
   "@material-ui/lab@^4.12.3":
     dependencies:
       "@material-ui/types": ^5.1.0
-  "@types/express-serve-static-core@*":
-    peerDependencies:
+  "@types/http-proxy@*":
```
freeqaz, line 27: These changes are to silence warnings on startup from Yarn PNP. :sob:

factoidforrest (approved):
Looks good to me. The one thing I don't think you should do is take the main backend off of Express 5. I think I was using something in there and I don't think it'll necessarily show until later how it's broken. I can't remember why but unless you have a specific reason, I'd ask you don't take it off of the beta.

.yarnrc.yml
factoidforrest: Holy fuck you did it

PR #930, opened 2022-10-12
ingest npm registry as worker

breadchris:
LunaTrace will not access npm's registry directly, instead it requests package manifest data through a proxy. The proxy accesses replicated npm information from the database.

The npm replicator runs independently of this flow and is constantly watching or change events from the npm couchdb event stream.


lunatrace/cli/gql/upsertRevision.graphql
```diff
@@ -0,0 +1,5 @@
+mutation UpsertRevisions($revisions: [npm_revision_insert_input!]!) {
```
factoidforrest, line 1: dead code

lunatrace/bsl/hasura/metadata/databases/lunatrace/tables/npm_revision.yaml
```diff
@@ -0,0 +1,34 @@
+table:
```
ajvpot, line 1: If hasura isn't being used any more we should probably remove this so it's not exposed in gql schema

PR #943, opened 2022-10-20
Fix typo on StringSubstitutor in text4shell blog

youribonnaffe:
It looks like there a typo on StringSubstitor mentioned at some point, it should be StringSubstitutor instead.


freeqaz:
Thanks for fixing this typo, @youribonnaffe!

PR #946, opened 2022-10-21
Datadog proxy CORS headers

freeqaz:
This makes the proxy work in production. I don't know why it was working for me in dev, but whatever. I tested against production and it works now. (this is already deployed)


factoidforrest (approved):
word

ajvpot (approved):

PR #957, opened 2022-10-26
add ref to releases from builds

ajvpot:

### DIFF
--- a/lunatrace/bsl/hasura/migrations/lunatrace/1666796296711_set_fk_public_builds_release_id/up.sql
+++ b/lunatrace/bsl/hasura/migrations/lunatrace/1666796296711_set_fk_public_builds_release_id/up.sql
@@ -1,0 +1,5 @@
+alter table "public"."builds"
+  add constraint "builds_release_id_fkey"
+  foreign key ("release_id")
+  references "package"."release"
+  ("id") on update restrict on delete restrict;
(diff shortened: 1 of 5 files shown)


lunatrace/bsl/hasura/migrations/lunatrace/1666796296711_set_fk_public_builds_release_id/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."builds"
```
factoidforrest, line 1: maybe Im wrong but couldnt all of this and these two migrations just be 

```
ALTER TABLE "public"."builds" ADD COLUMN "release_id" uuid NULL REFERENCES package.release ON UPDATE restrict ON DELETE restrict;
```

Also if we stick to capital postgres that's probably for the win, that's what everything else is in the project more or less. At least everything the rest of us have wrote.

factoidforrest (approved):
I wonder if we can make this a simpler migration. Maybe theres some reason not, just asking. 

Also capital letters for sql

lunatrace/bsl/hasura/migrations/lunatrace/1666796296711_set_fk_public_builds_release_id/up.sql
ajvpot: Makes sense. These were generated from hasura but I'll format them using our standard and merge them.

PR #963, opened 2022-10-29
Make empty vulnerabilities not clickable

factoidforrest:
As I feared, importing unpopulated vulnerabilities from github has resulted in a lot of empty vulnerabilities in the UI that render as broken/useless. As of this PR, they show in the "related vulnerabilities" table in the UI but are not clickable.

This fix adds the column "last_fetched" to the vulnerabilities table (and also a created_at, which we would ideally always have).
Perhaps we should get in the habit of always creating this last_fetched column whenever an ETL job is involved, because not only is it a useful filter, but it's also very nice to show the user when the data was last synced. 

Also very helpful when we plan on inserting stubs to be fetched later. 

We are in a similar situation with the NPM package data, and in that case we have the "last_failed_fetch" and "last_successful_fetch" columns.


ajvpot:
Looks like there are some problems with generated files

lunatrace/bsl/frontend/src/pages/vulnerabilities/detail/EquivalentVulnerabilitiesList.tsx
```diff
@@ -40,6 +42,16 @@ export const AffectedPackagesList: React.FC<{ relatedVulns: VulnInfoDetails['equ
                 {relatedVulns.map(({ equivalent_vulnerability: relatedVuln }) => {
                   const severity = getCvssVectorFromSeverities(relatedVuln.severities);
 
+                  if (!relatedVuln.last_fetched) {
```
ajvpot, line 45: Should we just hide vulns in this case? I think vuln ingestion is an implementation detail we should not expose to the user.

ajvpot (changes requested):
looks good but needs generated files fixed

PR #968, opened 2022-11-01
make each manifest have a ref to a root node in the dep tree

ajvpot:


factoidforrest (approved):
approved with the understanding that release_version will hold the buildId for lack of a better UUID

PR #975, opened 2022-11-02
reorganize builds folder

factoidforrest:


lunatrace/bsl/backend-cdk/lib/constants.ts
```diff
@@ -18,3 +18,9 @@ export const commonBuildProps: AssetImageProps = {
     buildArgs: false,
   },
 };
+
+// All other environment configs should inherit from this shared one
+export const baseEnvironmentVars = {
+  DEPLOYMENT_ENVIRONMENT: 'production',
```
ajvpot, line 24:
```suggestion
  LUNATRACE_ENVIRONMENT: 'production',
```

ajvpot (changes requested):
Looks good except env var namespacing. I think we should use LUNATRACE_ prefix like the others.

lunatrace/bsl/backend-cdk/lib/constants.ts
factoidforrest: I forgot this was even in here. Why did I want this again? Oh right, the ingester logs are completely garbled by the progress bar. Anyway I couldn't figure out how to get this into the go config, can you own it Alex? I'll pull this from here

PR #979, opened 2022-11-05
Chains scroll

factoidforrest:
reasonable chain scrolling. It looks nice, it's using the react-perfect-scrollbar lib which is great looking and has good touch support and so on. 

In the future, ideally, we would render an actual tree here instead of chains, and in that case the chains would be a lot more reasonably sized in most cases. That's a lot of work, though.


lunatrace/bsl/frontend/src/pages/project/builds/details/vulnerable-packages/body/DepChains.tsx
```diff
@@ -14,8 +14,11 @@
 import React, { useState } from 'react';
 import { Badge, NavLink } from 'react-bootstrap';
 import { ChevronRight, ChevronsRight, Maximize2, Minimize2 } from 'react-feather';
+import PerfectScrollbar from 'react-perfect-scrollbar';
```
ajvpot, line 17: What does this library do that can't be accomplished with css overflow property?

breadchris: oh my god https://perfectscrollbar.com/

lunatrace/bsl/hasura/migrations/lunatrace/1667411530061_build_logs_cascade/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE public.build_log
+    DROP CONSTRAINT build_log_build_id_fkey;
+
+ALTER TABLE public.build_log
+    ADD FOREIGN KEY (build_id) REFERENCES builds
```
breadchris, line 5: doesn't this already reference builds? https://github.com/lunasec-io/lunasec/blob/master/lunatrace/bsl/hasura/migrations/lunatrace/1662408856453_create_build_log/up.sql#L20

breadchris (changes requested):

lunatrace/bsl/frontend/src/pages/project/builds/details/vulnerable-packages/body/DepChains.tsx
factoidforrest: great touch support, animations like bounce back, pretty CSS for the bars, controls over the bar size, etc. It works well and its two lines.

breadchris (approved):

PR #980, opened 2022-11-05
fix broken log statements

factoidforrest:
this is breaking prod I think


factoidforrest:
merging based on the general directive that we can merge small hotfixes autonomously

PR #981, opened 2022-11-06
update dependency tree to use graph

breadchris:
use graph library to handle dependency tree creation


factoidforrest:
I heard you talking to free that maybe this was 25x slower than our hash based index solution?

Would another option be to simply break apart our logic into smaller modules accross files? For instance, one module builds vulns, another one organizes vulns by package, another one handles the tree data? 

What we really need is separation of concerns to keep this maintainable.

PR #983, opened 2022-11-08
Use BuildID for sharding instead of date

ajvpot:
Switches SBOM bucket over to use first characters of uuid as shard keys instead of the date.

Previously, we were writing all snapshots for an org to the same shard because we were sharding by orgID+date. This would have given us a limit of 3500 SBOM PUTs per second per org. We're not hitting that any time soon but this should distribute the load more evenly.


factoidforrest:
3500 SBOM PUTs per second per org

I think we would probably never, ever hit that. Ever. I don't even think snyk would hit that. 

Still, fair game.

factoidforrest (approved):

PR #993, opened 2022-11-15
fix chains and master slowdown

factoidforrest:


lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
```diff
@@ -182,19 +189,22 @@ export default class VulnerabilityDependencyTree {
...
+      packageReleaseVulnerabilities.forEach((releaseVulnerability, index) => {
+        const existingRelease = vulnerableReleasesById.get(node.release_id);
+        const isFirstVulnOnNode = index === 0;
+
+        const upsertedRelease = createOrMergeVulnerablePackageRelease(
           existingRelease,
           node.release,
           releaseVulnerability,
-          vulnerableNodeChains,
+          isFirstVulnOnNode ? chains : [],
```
factoidforrest, line 201: this and a one line change in this function that amalgamates the chains are the only important changes in this pr

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/graph.ts
```diff
@@ -12,33 +12,33 @@
  *
  */
 export class Graph {
-  private forwardEdges: Map<string, string> = new Map();
-  private backwardEdges: Map<string, Set<string>> = new Map();
+  private childEdges: Map<string, string> = new Map();
```
breadchris, line 15: this should probably be called 'parentIdToChildId'

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/graph.ts
```diff
@@ -12,33 +12,33 @@
...
+  private parentEdges: Map<string, Set<string>> = new Map();
```
breadchris, line 16: this should probably be 'childIdToParentIds'

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/graph.ts
```diff
@@ -12,33 +12,33 @@
...
+      this.childToEdgeIds.set(childId, new Set([edgeId]));
     } else {
-      existingEdgesToChild.add(id);
+      existingEdgesToChild.add(edgeId);
     }
   }
 
-  public inboundNeighbors(child: string) {
-    return this.backwardEdges.get(child) || new Set();
+  public getParentEdges(childId: string) {
```
breadchris, line 37: this should be `getParentIdsOfChild`

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
```diff
@@ -55,28 +56,28 @@ export default class VulnerabilityDependencyTree {
 
     sourceManifests.forEach((manifest) => {
       manifest.child_edges_recursive?.forEach((edge) => {
-        const child = edge.child;
+        const node = edge.child;
```
breadchris, line 59: I feel like keeping this called `child` is reasonable. It is consistent with how an `edge` is defined in a tree or acyclic graph.

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
```diff
@@ -141,7 +143,12 @@ export default class VulnerabilityDependencyTree {
   public getVulnerableReleases(): VulnerableRelease[] {
     const vulnerableReleasesById: Map<string, VulnerableRelease> = new Map();
 
-    this.vulnerableNodes.forEach((nodeId) => {
+    const previouslyLoopedNodeIds: string[] = [];
+    this.vulnerableNodeIds.forEach((nodeId) => {
+      if (previouslyLoopedNodeIds.includes(nodeId)) {
+        throw new Error('dupe node looped');
```
breadchris, line 149: either remove this or improve the message that is thrown and in the places that `getVulnerableReleases` is called, wrap in a try catch.

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
breadchris: I don't think we need `isFirstVulnOnNode` anymore now that `existingRelease` is properly set.

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
```diff
@@ -182,19 +189,22 @@ export default class VulnerabilityDependencyTree {
...
+          isFirstVulnOnNode ? chains : [],
           devOnly,
           triviallyUpdatable
         );
-        vulnerableReleasesById.set(node.release_id, updatedRelease);
+        vulnerableReleasesById.set(node.release_id, upsertedRelease);
       });
     });
-    return [...vulnerableReleasesById.values()];
+    return Array.from(vulnerableReleasesById.values());
```
breadchris, line 208: nice this looks better

package.json
```diff
@@ -125,6 +125,7 @@
     "node-gyp": "~9.0.0"
   },
   "resolutions": {
-    "snyk-nodejs-lockfile-parser": "portal:/home/forrest/code/nodejs-lockfile-parser"
+    "snyk-nodejs-lockfile-parser": "portal:/home/forrest/code/nodejs-lockfile-parser",
```
breadchris, line 128: need to fix these

breadchris (changes requested):

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
factoidforrest: yeah we still do, we can only do this once per node and the same node may have multiple vulnerabilities. Doing it again for each vulnerability would produce duplicate chains on the release

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/graph.ts
factoidforrest: yeah, i could see that luckily. Luckily when you go to use it reads like `parentEdges(childId)` which is pretty explanatory

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
factoidforrest: actually, it's not. There is no such thing as a "child" in the database, that's simply the name of the relation. The name of the object in the database is actually "ManifestDependencyNode".. hence, node

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/graph.ts
factoidforrest: ah yeah right! no edge id

lunatrace/README_DEV_SETUP.md
```diff
@@ -83,8 +83,13 @@ You'll need to format those values into an env file at `$REPO/lunatrace/dev-cli/
 S3_SBOM_BUCKET=xxx
 S3_MANIFEST_BUCKET=xxx
 QUEUE_NAME=xxx
+
+
+LUNATRACE_GRAPHQL_SERVER_URL=http://localhost:8080/v1/graphqlGGG
```
ajvpot, line 88: `graphqlGGG`?

lunatrace/bsl/backend/src/models/vulnerability-dependency-tree/index.ts
freeqaz: `childNode` might be easier to grok. By itself, I'd find `node` ambiguous.

PR #1007, opened 2022-11-29
database migration for environmental cvss

factoidforrest:


factoidforrest:
I guess we could change this to be "risk_adjustment" to be in line with the UI but we are probably going to change the UI again so..yolo

PR #1008, opened 2022-11-29
Add media to EPSS post, title change

ajvpot:
Adds youtube video to epss post, fits our theme of having media at the top of post.

Changes the title to remove reference to EPSS as a replacement for CVSS. EPSS depends on the CVSS information and they have different use cases so I don't think we should call it a direct replacement.

Replacement was not mentioned in the post except the title

### DIFF
--- a/docs/blog/2022-11-21-epss.mdx
+++ b/docs/blog/2022-11-21-epss.mdx
@@ -21,6 +21,8 @@ authors: [forrest, alex, free]
   ~
 -->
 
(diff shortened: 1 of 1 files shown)


docs/blog/2022-11-21-epss.mdx
```diff
@@ -63,7 +65,7 @@ Security Information and Event Management (SIEM) tools.**)
 * Count of how many references are listed in the CVE
 * Published Exploit code in any of: Metasploit, ExploitDB and/or Github
 * Security Scanners: Jaeles, Intrigue, Nuclei, sn1per
-* CVSS v3 vectors in the base score (not the score or any subscores) as published in the National Vulnerability Database (NVD)
```
ajvpot, line 66: This was redundant, NVD doesn't publish temporal or environmental modifiers.

PR #1015, opened 2022-12-01
callsites are stored in the database

breadchris:
It wasn't working for a while, and then I found what was wrong https://www.twitch.tv/videos/1667123403


breadchris:
I have read the CLA Document and I hereby sign the CLA

PR #1021, opened 2022-12-05
Builds create releases

ajvpot:
Builds create a root node in the merkle tree that references a release that represents the build.
The package associated with the root node has a unique custom registry based on the project ID and is marked as internal.
A new release is created for each build even if the version number is not changed so we can store the blob url.

### DIFF
--- a/lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
+++ b/lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
@@ -231,7 +231,7 @@ async function insertNodesToDatabase<Ext>(
         packageData.name || '',
         packageData.packageManager,
         // top nodes should use a custom namespace per org.
-        isTopLevel ? 'lunatrace-internal-project-' + projectId : packageData.customRegistry,
+        isTopLevel ? `lunatrace-internal-project-${projectId}` : packageData.customRegistry,
         // top nodes don't have a parent and are internal.
         !isTopLevel
       );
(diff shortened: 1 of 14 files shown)


lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
```diff
@@ -108,14 +112,19 @@ export async function collectPackageGraphsFromDirectory(repoDir: string): Promis
       // This value is set to false by the library when there are zero dev dependencies
       const prodOrDevLabel = pkgTree.hasDevDependencies ? 'dev' : 'prod';
 
-      const pkgDependenciesWithGraphAndMetadata = Object.values(pkgTree.dependencies).map((pkg) => {
+      // Add the unique id to the version of the root node if requested so it has a unique key in the db.
+      if (rootUniqueId) {
+        pkgTree.version = `${pkgTree.version}-${rootUniqueId}`;
+      }
+
+      const pkgDependenciesWithGraphAndMetadata = [pkgTree, ...Object.values(pkgTree.dependencies)].map((pkg) => {
```
breadchris, line 120:
```typescript
const treeWithMetadata = Object.values(pkgTree.dependencies)].map((pkg) => { ... });
const pkgDependenciesWithGraphAndMetadata = [pkgTree, ...treeWithMetadata];
```

breadchris (approved):
one style change

lunatrace/bsl/backend/src/snapshot/node-package-tree.ts
ajvpot: That's not what this does. It adds the root node to the list and maps over that.

PR #1051, opened 2022-12-16
use global cache

ajvpot:
since we have to yarn install to build murmurhash anyway now we might as well use the global cache.
This will allow us to reduce the repo size so go users can import our code without downloading a bunch of dead js code.

CI dep preparation time will be doubled until we use BFG to clean the repo, then should be practically the same.


factoidforrest (changes requested):
This might be a good one to chat about before merge. I feel like there are security guarantees to using immutable that we sort of need, unless I'm not understanding this correctly.

factoidforrest (approved):
noice

factoidforrest:
yarn install in CI seems to only take a minute, thats perfectly fine!

Issue #998, opened 2022-11-18
Environmental CVSS


factoidforrest:
Steps along our environmental journey:

1.) Implement a conditional CVSS system that runs at a project level of granularity. Ask a question or two about peoples projects and thats it.

2.) Implement folder level controls and then use the conditional CVSS on that

PR #1055, opened 2022-12-16
Environmental cvss

factoidforrest:
#998


lunatrace/bsl/frontend/src/pages/project/dashboard/DefaultBranchSummary.tsx
```diff
@@ -33,7 +33,7 @@ export const DefaultBranchSummary: React.FC<DefaultBranchSummaryProps> = ({ proj
   const defaultBranchBuilds = project.default_branch_builds;
   const defaultBranchBuild = defaultBranchBuilds ? defaultBranchBuilds[0] : null;
   const latestBuildAnyBranch = project.builds[0];
-  // Prefer a build on the default branch if available, otherwise use a build from any branch just so we can show something here
+  // Prefer a build on the default branch if available, otherwise use a builddev:sass from any branch just so we can show something here
```
ajvpot, line 36: bad sed?

lunatrace/bsl/frontend/src/pages/project/risk-adjustment/RiskAdjustmentMain.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const RiskAdjustmentMain: React.FC<FiltersProps> = ({ project }) => {
+  return (
+    <Container>
+      <Row>
+        <Col md="auto">
+          <h2>Risk Adjustment</h2>
+        </Col>
+        <Col md className="mt-md-1">
+          <p>
+            <Info size="1.2em" className="mb-1 me-2" /> LunaTrace uses these simple settings to adjust vulnerability
```
ajvpot, line 36: I think we could replace this copy with

"LunaTrace uses this information to prioritize vulnerabilities that pose the most risk to your infrastructure. If you are unsure about a specific setting, leave it blank."

We probably don't need the note about how effects are in development.

lunatrace/bsl/hasura/migrations/lunatrace/1669922910149_add_root_to_folder_options/up.sql
```diff
@@ -0,0 +1,85 @@
...
+
+ALTER TABLE public.cvss_environmental_adjustment
+    ADD COLUMN group_name TEXT;
+ALTER TABLE public.cvss_environmental_adjustment
+    ALTER COLUMN group_name SET NOT NULL;
+
+--  internet exposure
+--  TODO: when implementing this in the backend, make sure these internet settings can only LOWER a vulnerabilities value and not raise it, since these are overrides
+INSERT INTO public.cvss_environmental_adjustment (group_name, name, attack_vector)
+VALUES ('internet', 'internet', 'N'),
```
ajvpot, line 16: should probably give these known uuids because they will be referenced in code (i think)

factoidforrest: We discussed that.  It definitely occured to me. We landed on using the names as the way to identify them in code. 

The reasons we went with that 
1.) We want to have this feature be easily expandable with user-entered values in the future. Those obviously can't be based around known IDs and need to be more dynamic.
2.) Manually creating special known UUIDs feels so weird. Mainly, it's 1 though

lunatrace/bsl/frontend/src/pages/project/risk-adjustment/FolderList.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  useEffect(() => {
+    setNonRootFolders(savedNonRootSettings);
+  }, [project.project_folder_settings]);
+
+  const [triggerCustomMutation, _customMutationResult] = api.endpoints.customProjectUpdateMutation.useMutation();
+
+  const changePrecedence = (oldPrecedence: number, newPrecedence: number): void => {
+    const reorderedSettings = reorder(nonRootFolders, oldPrecedence - 1, newPrecedence - 1);
+    const gqlMutationStrings = reorderedSettings.map((setting, index) => {
```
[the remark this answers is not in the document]
factoidforrest: Yeah, I wanted this to go as one request. There is no way in RTK that I could figure out to do this without parameterizing the entire query body like I did here. It seemed like a necessary evil. 

One option would be to hide this weirdness in another file as a helper, that way its not in the component.

PR #1066, opened 2022-12-28
New backend config

factoidforrest:
This changes the env var config loader in the node backend to a much more generic solution. It now throws when a required env var is missing at START TIME rather than run time.


factoidforrest:
YESSSSS IT PASSED

lunatrace/bsl/backend/src/config/index.ts
```diff
@@ -0,0 +1,164 @@
...
+} from '../types/config';
+import { tryParseInt, validateBooleanString } from '../utils/parse';
+
+import { envVars } from './load-environment-vars';
+
+export { envVars } from './load-environment-vars';
+export const isProduction = envVars.NodeEnv === 'production';
+
+// The below loaders build configs for various parts of the app.
+// TODO: There is no reason in node to load configs like this in NodeJS, its just extra code and not real DI.
```
breadchris, line 38: The reason the configuration was written like this was to allow dependency injection to be possible. I do not think environment variables should be referenced directly from a global context.

lunatrace/bsl/backend/src/config/load-environment-vars.ts
```diff
@@ -0,0 +1,171 @@
...
+  // AND we are in production and the default value is not defined.
+  // then throw an error
+  if (!envVar && !backupVar && defaultValue === undefined && process.env.NODE_ENV !== 'test') {
+    missingEnvVars.push(`${envVarKey}${backupEnvVarKey ? ` and backup var ${backupEnvVarKey}` : ''}`);
+    // throw new Error(`Missing ${envVarKey} env var ${backupEnvVarKey ? `and backup var ${backupEnvVarKey}` : ''}`);
+  }
+  return envVar || backupVar || (defaultValue as string);
+};
+
+const commonEnvVarKeys = {
```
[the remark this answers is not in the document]
breadchris: Loading environment variables like this means that the workers and the backend need to all have the same environment variables in production (which they currently don't have).

lunatrace/bsl/backend/src/config/load-environment-vars.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type EnvVar<S extends z.Schema> = {
+  varKey: string;
+  backupVarKey?: string;
+  defaultValue?: z.infer<S>;
+  castTo: S;
+};
+
+// Takes an env var key object from "common-env-vars.ts", reads it from the environment, and returns a scalar or throws if it cannot
+function checkEnvVar<E extends EnvVar<z.Schema>>(varConf: E): z.infer<E['castTo']> {
```
ajvpot, line 31: This is incredibly confusing

ajvpot (approved):

lunatrace/bsl/backend/src/config/load-environment-vars.ts
breadchris: can you elaborate on why you find this “incredibly confusing”?

factoidforrest: I mean, i agree with alex. This is a small amount of really confusing code. 

In the future, we should look first for libraries when we go to do things like this. It will save us a lot of time.

PR #1070, opened 2023-01-02
Add EPSS Ingestor command to ingest-worker

freeqaz:
Detailed Summary:

- This commit adds a new epss command to the ingest-worker.
- The epss command includes an "ingest" subcommand which allows the user to update EPSS scores.
- The epss command is implemented using the clifx library and includes a NewCommand function to create the cli.Command struct.
- The epss command has a dependency on an EPSSIngester interface which is provided through the fx dependency injection library.
- The epss command logs informative messages when starting and completing the ingestion process.

(Thank ChatGPT for the git commit message lol)

The SQL I wrote is the fastest that I could figure out. The first implementation took >10 minutes and this one takes about 30 seconds now.

It turns out that query for the IDs ahead of time and then running an "UPDATE" for that is the fastest way. (worth noting)


lunatrace/bsl/ingest-worker/pkg/epss/ingester.go
```diff
@@ -0,0 +1,151 @@
...
+    Isolation: sql.LevelReadUncommitted,
+  })
+  if err != nil {
+    return err
+  }
+  defer tx.Rollback()
+
+  // Get all CVEs and GHSAs from the database
+  cvePairs, err := tx.QueryContext(ctx, `
+    SELECT v1.source_id, v2.source_id FROM vulnerability.equivalent e
```
factoidforrest, line 75: im glad this works for now, but there must be a real library we can use for this in the future. passing the raw sql strings is starting to look clunky, although god knows ive done it a lot myself

factoidforrest: In node there are quite a few query builder libraries. There are also libraries which will read in .sql files and template stuff into them. That has the advantage of code separation, and it will allow the IDE to work some autocomplete magic on the SQL. We might consider that in the future

factoidforrest (approved):

ajvpot:
please run go fmt on your changes, they don't match the [language style](https://go.dev/doc/effective_go#formatting) specifically use tabs for indentation and there is no limit on line length.

ajvpot (changes requested):
please run go fmt on your changes, they don't match the [language style](https://go.dev/doc/effective_go#formatting) specifically use tabs for indentation and there is no limit on line length.

lunatrace/bsl/ingest-worker/cmd/ingestworker/vulnerability/vulnerability.go
```diff
@@ -62,22 +65,52 @@ func NewCommand(p Params) clifx.CommandResult {
...
-							Msg("Updating CWEs")
-						err := p.CWEIngester.Ingest(ctx.Context)
-						if err == nil {
+						runIngestion := func() error {
+							log.Info().
+								Msg("Updating CWEs")
+
+							err := p.CWEIngester.Ingest(ctx.Context)
+
+							if err != nil {
```
ajvpot, line 74: Please collect errors instead of discarding them. Returning an error from this func should cause the tool to exit with nonzero exit code which will trigger monitoring if it fails.

lunatrace/bsl/ingest-worker/pkg/epss/fetch.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// EpssVulnerability has a CVE, Epss, and Percentile
+type EpssVulnerability struct {
+  Cve        string
+  Epss       float64
+  Percentile float64
+}
+
+func FetchEpssScores() (epssScores []EpssVulnerability, err error) {
```
ajvpot, line 32: nit: I would move these into the function and only allocate the epssScores array after we've read the CSV. We can save allocations by [specifying the capacity of the slice](https://github.com/uber-go/guide/blob/master/style.md#prefer-specifying-container-capacity) when we create it.

lunatrace/bsl/ingest-worker/pkg/epss/fetch.go
```diff
@@ -0,0 +1,120 @@
...
+  if err != nil {
+    fmt.Printf("Error creating gzip reader: %v\n", err)
+    return
+  }
+  defer gzipReader.Close()
+
+  // Create a new CSV reader
+  csvReader := csv.NewReader(gzipReader)
+
+  // We have to set this or else we get a "wrong number of fields" error for some reason
```
ajvpot, line 69:
```
	// FieldsPerRecord is the number of expected fields per record.
	// If FieldsPerRecord is positive, Read requires each record to
	// have the given number of fields. If FieldsPerRecord is 0, Read sets it to
	// the number of fields in the first record, so that future records must
	// have the same field count. If FieldsPerRecord is negative, no check is
	// made and records may have a variable number of fields.
```
[Godoc](https://pkg.go.dev/encoding/csv#:~:text=number%20of%20fields.-,FieldsPerRecord,-int%0A%0A%09//%20If%20LazyQuotes) can usually help when you are seeing weird semantic behavior of a package. The error comes from there being a varying number of fields in the first few rows of the file.

freeqaz: It's because the first line of the CVS is a weird format. It looks like I just needed to set `csvReader.Comment = '#'` to fix that. Thanks :pray:

lunatrace/bsl/ingest-worker/pkg/epss/ingester.go
freeqaz: Screenshot_2023-01-03_15-54-53

The SQL autocomplete works great for me FWIW

PR #1100, opened 2023-02-01
Scaffold LunaTrace NPM CLI

freeqaz:
This is built around Arborist and will be the place where we eventually are able to generate "patches" for an arbitrary change.

My goal with this was to just create a basic command, in this case `show-tree`, which wraps Arborist and provides a basic "test case"
for us to verify that the library works.

There is a basic fixture that I committed here -- an arbitrary NPM project from my machine -- that we can use for testing.

In the future, I hope to add in the commands to actually modify the tree to "update" a package to a given version.


factoidforrest (approved):
Looks great. We discussed maybe not needing the CLI functionality at all for the time being, instead using this directly as a JS import, but that's still good.  We could maybe pull the CLI stuff for now so it's not dead code, or leave it if you think we will someday do that. 

Did you want this as BSL in that case?  Just asking, doesn't really matter either way I don't think. 

Let's ship it.

PR #1106, opened 2023-02-06
Survey

factoidforrest:
moreinfo_003

Survey with open ended question boxes. Appears for new users directly after signup, and for all existing users who have not yet completed the survey.  Blocks use of the app for new users without projects (sidebar will be hidden by our current "wizard" system). Replaces access to the home page for existing users, but projects will still be accessible via sidebar.

Fields are required by html5 validation. stored in db as jsonb. Jsonb column size is limited at the DB level, but no other structural validation is done. I left this very open because questions may change and theres no real reason I know of for this data to be super contrete, we are a startup. Unless there is some jsonb postgres vuln or something. @freeqaz please let me know if you see any security concerns in this, would like different questions, etc. You (and Yiannis) will be the only ones actually using the data so let me know whatever you need! 

There is no data retrieval in the app as part of this PR. If you would like to view these through an admin panel instead of through prod hasura, let me know @freeqaz


freeqaz (approved):
I like that you chose JSONB for the backend storage part of this.

I think we'll want to tweak these questions (probably have a discussion about _what_ is most useful for us), but that should be easy enough given that this is JSONB :)

Looks good overall and we should land this!

PR #1137, opened 2023-02-22
Vulnerability Blog Post

breadchris:
This blog post outlines an analogy between holes in a boat and vulnerabilities in code. The analogy tries to capture the point that not all holes in your boat have to patched immediately.


factoidforrest:
@breadchris I took the liberty of merging master since it fixes the docs and a whole lot else. I was also able to update docusaurus to latest as part of this branch. woohoo. 

Your front mater has not been updated, you still have the title from the previous blog post. I think you saw that. We just need to fix that and then we can merge

factoidforrest (changes requested):
.

PR #1145, opened 2023-02-27
Test storing embeddings in postgres

breadchris:
Test out https://supabase.com/blog/openai-embeddings-postgres-vector to see if we can use it


lunatrace/bsl/hasura/migrations/lunatrace/1677516215552_vulnerable_reference_embeddings/up.sql
```diff
@@ -0,0 +1,56 @@
+CREATE EXTENSION vector;
+
+CREATE TABLE "vulnerability"."reference_content" (
```
factoidforrest, line 3: if this is a 1:1 with references, I'd usually say that it should just be part of the parent table. Thing is though, these columns are big, and while thats usually fine with a normal setup, im scared hasura might do some extremely stupid `SELECT * ...` queries. We should look into that

factoidforrest (approved):

breadchris (commented):
remove implementation of metadata embedder

PR #1153, opened 2023-03-17
add migrations for ML generated fields

factoidforrest:


lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    score integer NOT NULL,
+    summary text NOT NULL,
+    type text NOT NULL,
+    language text NOT NULL
+);
+
+-- Add a generated summary to the reference to make it easier for the LLM to choose what to read
+ALTER TABLE vulnerability.reference_content ADD COLUMN summary text NULL;
+
+ALTER TABLE package.package ADD COLUMN readme_text text NULL;
```
breadchris, line 21: A package already has `package.description` which often times the readme. Probably don't need to add another column to do the search.

lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
```diff
@@ -0,0 +1,22 @@
...
+ALTER TABLE package.package ADD COLUMN use_case_summary text NULL;
```
breadchris, line 22: What will this column contain that is different than description?

lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    vulnerability uuid NOT NULL references vulnerability.vulnerability,
+    code text NOT NULL,
+    score integer NOT NULL,
+    summary text NOT NULL,
+    type text NOT NULL,
+    language text NOT NULL
+);
+
+-- Add a generated summary to the reference to make it easier for the LLM to choose what to read
+ALTER TABLE vulnerability.reference_content ADD COLUMN summary text NULL;
```
breadchris, line 19: what is the difference between this column and https://github.com/lunasec-io/lunasec/blob/master/lunatrace/bsl/hasura/migrations/lunatrace/1678406466712_add_parsed_content_to_reference_content/up.sql#LL1C81-L1C81?

lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
factoidforrest: Why are we jamming the readme into that column? That isn't what I would expect, id expect it to have a short description. Also, often? When does it not?  Depending on ecosystem?

lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
factoidforrest: We discussed this a few times so I think you're familiar. By telling the LLM to summarize the use case specifically, and avoid all other descriptions, we get a much better vector proximity. It's not a general description, its just "what is this for"

lunatrace/bsl/hasura/migrations/lunatrace/1679066158985_add_ml_tables/up.sql
factoidforrest: We also discussed this the other day, this is a short one to two sentence description of the content, so that we can display it in a list for the LLM to choose what to read.

PR #1155, opened 2023-03-17
add langchain service

breadchris:
The langchain service is a grpc service that is callable with generated code

```go
	_, err = p.deps.LangChain.CleanWebpage(context.Background(), &gen.CleanWebpageRequest{
		Content:     normalContent,
		Description: ref.VulnDesc,
	})
```


factoidforrest (approved):
so awesome

PR #1157, opened 2023-03-19
Ml refinement

factoidforrest:


lunatrace/bsl/ingest-worker/pkg/vulnbot/vulnbot.go
```diff
@@ -79,7 +90,8 @@ func (v *vulnbot) messageHandler(ctx context.Context, info discordfx.MessageInfo
 			log.Error().Err(err).Msg("error processing message")
 			return
 		}
-		_, err = s.ChannelMessageSend(m.ChannelID, resp.Response)
+		// TODO: make this also show the intermediate steps in a collapsed box in discord (resp.IntermediateSteps). Bonus points if we can figure out how to preserve coloring
```
factoidforrest, line 93: @breadchris do you know how to do this?

breadchris: yeah we are going to have to make the proto call async https://grpc.io/docs/languages/python/basics/#simple-rpc-1 which isn't that difficult. def possible

